PR #55, opened 2026-03-01
feat: add background goroutine for scheduled audit log purge

### LINKED ISSUE #35: Set up scheduled audit log purge in Go app
## Context

We now have `db.PurgeExpiredAuditEvents()` (added in the data retention compliance work) which deletes audit events older than each user's plan retention window:
- Free tier: 7 days
- Paid tier: 90 days

The function is implemented and tested but not yet called on a schedule. Rather than running a separate cron service, we should run it as an in-process background goroutine in the Go app.

## Implementation

- [ ] Add a background goroutine in `main.go` that runs `db.PurgeExpiredAuditEvents()` on a ticker (e.g. every hour)
- [ ] Start the goroutine after database connection is established
- [ ] Use a `context.Context` derived from the server's shutdown context so the goroutine stops cleanly on `SIGTERM`
- [ ] Log the number of purged rows on each run (and any errors)
- [ ] Consider making the interval configurable via `AUDIT_PURGE_INTERVAL` env var (default: `1h`)
- [ ] Ensure the purge is safe to run concurrently on multiple Fly.io instances (the `DELETE...USING` query is idempotent — multiple instances running it simultaneously is fine, just redundant work)

## Notes

- On Fly.io with auto-stop/start, the goroutine only runs while the machine is awake — this is acceptable since the purge doesn't need to be real-time
- The purge query is efficient: it joins `audit_events` → `subscriptions` → `plans` and uses the plan's `audit_retention_days` column
- No migration needed — the function already exists in `db/data_retention.go`

chiedo:
Run db.PurgeExpiredAuditEvents() on a ticker (default: 1 hour) to
automatically clean up audit events past each user's plan retention
window. The goroutine starts after database connection, runs an
initial purge immediately, then repeats on the configured interval.

- Configurable via AUDIT_PURGE_INTERVAL env var (Go duration format)
- Uses context cancellation for clean shutdown on SIGTERM
- Logs purged row count on each run and reports errors to Sentry
- Safe for multi-instance (the DELETE query is idempotent)

Closes #35

https://claude.ai/code/session_012TeGpXWaRUBouTptdtUayW


chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedoclaude:
## 🤖 Watch Session Summary

### Changes Made
- **`feat: add background goroutine for scheduled audit log purge`** (`a6cbc87`) — Initial implementation: background goroutine running `db.PurgeExpiredAuditEvents()` on a configurable ticker (default 1h), with context-based clean shutdown, Sentry error reporting, and `AUDIT_PURGE_INTERVAL` env var support.
- **`fix: address review feedback on audit purge goroutine`** (`7591a05`) — Return done channel from `startAuditPurge` so shutdown waits for goroutine exit; skip Sentry for context cancellation errors; pass logger to `auditPurgeInterval` and include parse error; add 10 table-driven tests.
- **`fix: harden audit purge — min interval, shutdown timeout, log tuning`** (`d19bcbc`) — Enforce 1-minute minimum interval to prevent resource exhaustion; add 5s timeout on shutdown wait; use Debug level for zero-row purges; include env var name in startup log.
- **`refactor: extract audit purge to background.go for maintainability`** (`7e85baa`) — Move audit purge functions from main.go (610→539 lines) to background.go (82 lines) with corresponding test file.
- **`docs: document AUDIT_PURGE_INTERVAL in README and .env.example`** (`600a113`) — Add env var to production environment variables table and .env.example with retention window description.
- **`test: add goroutine lifecycle and error path tests for audit purge`** (`c5d69db`) — Cover startAuditPurge done channel lifecycle, runAuditPurge context cancellation vs real errors, and log level verification (Debug for 0 rows, Info for >0 rows). All pass with `-race`.

### Decision Log

#### ✅ Implemented
- Copilot suggested checking `context.Canceled`/`context.DeadlineExceeded` to avoid noisy Sentry events on shutdown → implemented via `ctx.Err() != nil` check in `7591a05`
- Copilot suggested returning a done channel / WaitGroup for clean goroutine shutdown → implemented done channel pattern in `7591a05`
- Copilot suggested using injected logger (not `slog.Warn`) and including parse error → fixed in `7591a05`
- Copilot suggested adding unit tests for `auditPurgeInterval` → added 10 table-driven tests in `7591a05`
- chiedobot asked for DX/UX improvements → added Debug-level for zero-row purges, env var name in startup log in `d19bcbc`
- chiedobot asked for security review → added 1-minute minimum interval floor and 5s shutdown wait timeout in `d19bcbc`
- chiedobot asked for maintainability/DRY/file splitting → extracted to background.go in `7e85baa`
- chiedobot asked for documentation improvements → added to README and .env.example in `600a113`
- chiedobot asked for previously-skipped harder tests → added goroutine lifecycle + mock-based error path tests in `c5d69db`

#### ❌ Declined
- None

#### ⚖️ Judgment Calls
- Chose `ctx.Err() != nil` over `errors.Is(err, context.Canceled)` for the cancellation check — simpler and covers both cancellation and deadline exceeded in one check
- Used done channel pattern instead of `sync.WaitGroup` — sufficient for a single goroutine and avoids the overhead of tracking multiple goroutines when there's only one
- Set minimum interval at 1 minute — low enough for testing/staging, high enough to prevent accidental tight loops

#### ❓ Open Questions
- None

---
*Watch session ended after 5 minutes of inactivity. Processed 8 comments across 9 poll cycles.*

Issue #90, opened 2026-03-01
Add Stripe connector (API key auth)


chiedoclaude:
## Summary

Add a built-in Stripe connector using the Stripe REST API. Stripe uses **secret API keys** — no OAuth required (Stripe Connect OAuth is for platform marketplaces, not needed here).

**Auth type:** `api_key` (secret key, `sk_live_...` or `sk_test_...`)
**Credential setup:** https://docs.stripe.com/keys

## Why Tier 2

Financial operations are high-value automation targets. Stripe powers payments for millions of businesses. Agents that can create invoices, issue refunds, and manage subscriptions unlock billing and finance workflow automation. Higher risk level but high ROI.

## Actions to Ship

| Action Type | Name | Risk | Key Parameters | Description |
|---|---|---|---|---|
| `stripe.create_customer` | Create Customer | low | `email`\*, `name`, `description`, `phone`, `metadata` (object — key/value pairs) | Create a new customer record — foundational for all other Stripe operations |
| `stripe.create_invoice` | Create Invoice | medium | `customer_id`\*, `description`, `due_date` (Unix timestamp), `auto_advance` (bool, default true — auto-finalize), `line_items[]` (array of `{description, amount, quantity}`), `metadata` | Create and optionally send an invoice |
| `stripe.issue_refund` | Issue Refund | high | `payment_intent_id`\* or `charge_id`\*, `amount` (cents — partial refund, omit for full), `reason` (duplicate/fraudulent/requested_by_customer), `metadata` | Refund a charge or payment intent — **high risk: moves real money** |
| `stripe.list_subscriptions` | List Subscriptions | low | `customer_id`, `status` (active/past_due/canceled/all), `price_id`, `limit` (default 10, max 100) | List subscriptions, optionally filtered by customer or status |
| `stripe.create_payment_link` | Create Payment Link | medium | `line_items[]`\* (array of `{price_id, quantity}`), `after_completion` (redirect URL or hosted page), `allow_promotion_codes` (bool), `metadata` | Create a shareable payment link for one-time or recurring purchases |
| `stripe.get_balance` | Get Balance | low | *(none)* | Retrieve the current account balance — useful for agents monitoring cash flow |

\* = required parameter

### Phase 1 — Connector scaffold *(sequential, ~1 day)* ✅

Must be completed first. Stripe has unique request encoding (form-encoded, not JSON).

- [x] Create `connectors/stripe/` package
[...]

PR #198, opened 2026-03-06
Add Stripe connector scaffold (Phase 1)

chiedoclaude:
## Summary

Implements Phase 1 of the Stripe connector (#90) — the foundational scaffold that all 6 actions will build on in Phase 2.

- **`StripeConnector` struct** with shared HTTP client, base URL, `newForTest()` constructor
- **`formEncode()` helper** for Stripe's bracket-notation form encoding (`metadata[key]=value`, `line_items[0][amount]=1000`) — handles nested objects, arrays, and flat values
- **Shared `do()` helper** — Bearer auth, `application/x-www-form-urlencoded` POST bodies, query params on GET, timeout/cancellation handling
- **`ValidateCredentials()`** — validates `sk_live_`/`sk_test_`/`rk_live_`/`rk_test_` prefixes
- **Error mapping** from Stripe error types → connector error types (`authentication_error` → AuthError, `invalid_request_error` → ValidationError, `rate_limit_error` → RateLimitError, `card_error`/`api_error` → ExternalError)
- **`Idempotency-Key` support** via `deriveIdempotencyKey()` — deterministic SHA-256 hash of action type + parameters (not random UUID) to preserve retry safety
- **29 tests** covering form encoding, credential validation, error mapping, HTTP lifecycle, and idempotency

### Design decisions

- **`do()` takes `map[string]string`** (pre-flattened params) rather than `map[string]any` — keeps the HTTP layer simple. Actions call `formEncode()` on their typed params before passing to `do()`.
- **Deterministic idempotency keys** instead of random UUIDs — a random UUID generated per `Execute()` call would defeat the purpose on retries. SHA-256 of action type + parameters ensures the same request always produces the same key.
- **Restricted key prefixes** (`rk_live_`/`rk_test_`) accepted in addition to standard `sk_` prefixes — these are Stripe's restricted API keys which are commonly used for scoped access.

## Test plan

- [x] All 29 tests pass (`go test ./connectors/stripe/... -v`)
- [x] Go compilation succeeds (`go build .`)
- [ ] Phase 2 actions will validate the scaffold works end-to-end

https://claude.ai/code/session_01YRRj5mxaqfrhsLkT4Rcc2T


chiedobot (commented):

chiedoclaude (commented):
Applied three DX improvements in 2bacdf9:

1. **`doGet()`/`doPost()` convenience methods** — Phase 2 action devs won't need to deal with the 7-param `do()` signature. `doPost()` auto-derives deterministic idempotency keys, so actions get retry safety for free.

2. **Pinned `Stripe-Version` header** (`2025-12-18.acacia`) — prevents silent breaking changes when Stripe releases new API versions. Version is a constant so it's easy to bump deliberately.

3. **Error codes in messages** — Stripe's `code` field (e.g., `card_declined`, `expired_card`) is now included in error messages, making it easier to debug and potentially handle specific errors at the action level.

33 tests pass.

chiedobot (commented):

chiedoclaude (commented):
Fixed two security issues in 90c5c8c:

1. **Unbounded response body read** — `io.ReadAll(resp.Body)` had no size limit. Added `io.LimitReader` with a 4 MB cap. This prevents memory exhaustion if the upstream returns an unexpectedly large response (malicious or misconfigured).

2. **Raw response body in error messages** — When Stripe returns a non-JSON error, the raw body was used as the error message with no length limit. This could leak sensitive data in logs or create oversized log entries. Now capped at 512 bytes with truncation.

Note: the existing Slack and GitHub connectors have the same `io.ReadAll` pattern without a limit. That's worth a separate hardening pass but is outside the scope of this PR.

35 tests pass.

chiedobot (commented):

chiedoclaude (commented):
Applied maintainability improvements in 39d68f8:

1. **Split `response.go`** — extracted `checkResponse`, `stripeError` into their own file, matching the GitHub connector's file structure. `stripe.go` dropped from 346 → 275 lines.

2. **Fixed naming inconsistency** — `livKeyPrefix` → `liveKeyPrefix` (was inconsistent with `rLiveKeyPrefix`).

3. **Extracted `validKeyPrefixes` var** — avoids rebuilding the slice on every `hasValidPrefix()` call.

4. **Simplified `encodeParams`** — removed redundant manual key sorting since `url.Values.Encode()` already sorts alphabetically.

5. **Race detector clean** — `go test -race` passes. All tests use `t.Parallel()` with fresh state per test, no shared mutable state.

Nothing else stood out as a real concern. The code follows established connector patterns, credentials are properly handled via the `Credentials` type (redacted in logs/JSON), and the form encoding uses `url.Values` for proper URL encoding.

chiedobot (commented):

chiedoclaude (commented):
Added documentation in 600d963:

1. **Stripe connector README** — comprehensive guide covering credentials, error mapping, form encoding, idempotency, API version pinning, file structure, testing, and instructions for adding Phase 2 actions. Follows the same template as GitHub and Slack connector READMEs.

2. **Updated `docs/creating-connectors.md`** — added a callout about non-JSON APIs (form-encoded), linking to the Stripe connector as a reference implementation. The existing guide assumes JSON throughout, so this helps developers who need to integrate with form-encoded APIs.

chiedobot (commented):

chiedoclaude (commented):
Honest audit — found two discrepancies, fixed in 34bc6eb:

**1. Test counts were off by 1 throughout.** I claimed 29 → 33 → 35 test functions across commits, but the actual counts are 30 → 34 → 36. I miscounted the original set by one. The deltas between commits were correct (added 4, then 2), but the base was wrong.

**2. `TestDo_LargeResponseTruncated` was misleading.** The test name and comment said it tested truncation of a response larger than `maxResponseBytes`, but the mock server only returned 11 bytes (`{"id":"ok"}`). It wasn't actually testing the limit — just that `LimitReader` doesn't break normal responses. Renamed to `TestDo_ResponseLimitReaderDoesNotBreakNormalResponses` with an accurate comment.

Everything else I claimed checks out against the code:
- All functions exist and work as described
- All error mappings are correct
- Idempotency key derivation is deterministic (SHA-256)
- `Stripe-Version` header is sent on every request
- Error codes are included in messages
- Response body reads are bounded
- `response.go` split matches the GitHub connector pattern
- Naming fix (`livKeyPrefix` → `liveKeyPrefix`) is done

chiedobot (commented):

chiedoclaude (commented):
Final polish pass applied in a2f968c — three things I'd been aware of but hadn't addressed:

**1. Missing compile-time interface check.** Both GitHub and Slack connectors use `var _ connectors.Connector = (*XConnector)(nil)` in their test files to catch interface drift at compile time. The Stripe connector only had runtime checks (`TestID`, `TestActions_ReturnsMap`). Added the compile-time check to match prior art.

**2. `truncate()` was not UTF-8 safe.** The previous implementation used `s[:maxLen]` which can split a multi-byte character (e.g., slicing `"日本語"` at byte 5 cuts the second kanji in half), producing invalid UTF-8 in error messages and logs. Fixed to walk runes and find a clean boundary. Added a test with multi-byte characters to verify.

**3. No test actually verified the LimitReader caps large responses.** The existing test (`TestDo_ResponseLimitReaderDoesNotBreakNormalResponses`) was renamed in the honesty audit because it only tested an 11-byte response. Added `TestDo_LargeResponseCappedByLimitReader` that sends a 4MB+ response and verifies the connector handles the truncated (non-JSON) body gracefully as an ExternalError.

Total: 38 test functions, all passing.

PR #363, opened 2026-03-08
Add Stripe Connect OAuth support alongside existing API key auth

### LINKED ISSUE #339: Add OAuth support for Stripe connector
## Summary

The Stripe connector currently uses API key auth. Stripe supports OAuth2 (Stripe Connect). We should add OAuth as an authentication option and make it the **default** in the UI.

## Requirements

- [ ] Add OAuth2 provider configuration for Stripe (via Stripe Connect)
- [ ] OAuth should be the **default authentication method** in the UI
- [ ] **Keep existing API key support** — do not remove it, add OAuth alongside it
- [ ] Users should be able to choose between OAuth and API key when configuring the connector
- [ ] Update frontend connector UI to present OAuth as the primary option with API key as an alternative

## Context

OAuth provides a better user experience (no manual token generation/rotation, granular scope control, automatic refresh). API keys remain useful for direct account access and users who prefer them.

chiedo:
## Summary
- Add OAuth2 via Stripe Connect as the default authentication method for the Stripe connector
- Keep existing API key support as a fallback — users can choose either method
- Register Stripe as a built-in OAuth provider with Connect OAuth endpoints
- Update execution layer with OAuth-first credential resolution and smart fallback
- Update frontend ConnectorCredentialsSection to render OAuth connect rows

## Test plan
- [x] Stripe connector tests pass (ValidateCredentials, do() with access_token and API key)
- [x] OAuth builtin tests pass (Stripe provider registered with correct endpoints/scopes)
- [x] Frontend ConnectorCredentialsSection tests pass (9/9)
- [x] TypeScript compiles with zero errors
- [x] All Go packages compile cleanly
- [ ] Manual: configure Stripe Connect OAuth credentials and verify OAuth flow works end-to-end
- [ ] Manual: verify API key fallback still works when no OAuth connection exists

Closes #339

<https://claude.ai/code/session_013kxDyH7gri48SGjgD2y2kg>

### DIFF
--- a/api/connector_execution.go
+++ b/api/connector_execution.go
@@ -41,25 +41,12 @@ func executeConnectorAction(ctx context.Context, deps *Deps, userID, actionType 
 		return nil, nil
 	}
 
-	// Check auth_type for this connector's required credentials.
-	reqCred, err := db.GetRequiredCredentialByActionType(ctx, deps.DB, actionType)
+	// Resolve credentials. Connectors that support multiple auth methods
+	// (e.g. Stripe with both OAuth and API key) are handled by trying OAuth
+	// first and falling back to static credentials.
+	creds, err := resolveCredentials(ctx, deps, userID, actionType)
 	if err != nil {
-		return nil, fmt.Errorf("look up required credential: %w", err)
-	}
-
-	var creds connectors.Credentials
-	if reqCred != nil && reqCred.AuthType == "oauth2" {
-		// OAuth2 path: resolve access token from oauth_connections.
-		creds, err = resolveOAuthCredentials(ctx, deps, userID, reqCred)
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		// Static credential path (api_key, basic, custom): existing behavior.
-		creds, err = resolveStaticCredentials(ctx, deps, userID, actionType)
-		if err != nil {
-			return nil, err
-		}
+		return nil, err
 	}
 
 	// Validate credentials before executing the action.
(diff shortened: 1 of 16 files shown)


chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedoclaude (commented):
**Security audit (re: review 3910159702):**

Found and fixed one real issue: `OAuthCredentialRow.handleConnect` used `fetch` with `redirect: "manual"`, but per the Fetch spec, opaqueredirect responses don't expose headers — `response.headers.get("Location")` always returns `null`. The "secure" code path was dead code; it always fell through to putting the token in the URL. Removed the dead code and simplified to match `ConnectedAccountsSection`'s existing pattern. (commit f5d06a2)

Other security findings (all clean):
- **Go connector**: `resolveKey()` properly prefers OAuth `access_token` over static `api_key`. Key prefix validation is sound.
- **Token handling**: OAuth tokens are stored encrypted in the vault. Token refresh has proper timeout (10s) and race-condition handling.
- **Input validation**: All Stripe action handlers validate required fields before making API calls. No SQL injection vectors.
- **Error messages**: No credential values are leaked in error responses.
- **CORS/auth**: Backend OAuth endpoints properly validate the bearer token before redirecting.

**Code quality audit (re: review 3910189386):**
- **DRY**: Provider labels already consolidated into `lib/oauth-providers.ts` (previous commit). No remaining duplication found.
- **File sizes**: All changed files are reasonable size. `ConnectorCredentialsSection.tsx` grew to ~340 lines with the new `OAuthCredentialRow` and `StaticCredentialRow` components, but each has a single clear responsibility.
- **Test race conditions**: All Go tests use `t.Parallel()` with per-test HTTP servers. Frontend tests use isolated mocks. No shared mutable state.
- **Future-proofing**: The dual-auth pattern (OAuth + API key) follows the same approach established by the Intercom connector. `resolveCredentialsWithFallback` in the execution layer handles the fallback generically.
- **Type safety**: All new TypeScript code respects `noUncheckedIndexedAccess`. No `any` or unsafe casts.

chiedobot (commented):

chiedoclaude (commented):
**Documentation improvements (re: review 3910208481):**

Added in commit 8691a24:

1. **`docs/oauth-setup.md`** — Added full Stripe OAuth setup section:
   - Stripe Connect setup steps (enable Connect, get client ID)
   - Redirect URI configuration
   - Environment variables (`STRIPE_CLIENT_ID`, `STRIPE_CLIENT_SECRET`) with descriptions
   - Scope explanation (`read_write`)
   - "How it works" explanation of the Connect OAuth flow
   - Added Stripe to the redirect URI troubleshooting list

2. **`connectors/stripe/README.md`** — Updated Credentials section:
   - Expanded from single API key row to a table showing both OAuth and API key methods
   - Documented `access_token` priority over `api_key`
   - Added link to the OAuth setup guide
   - Clarified that key prefix validation applies to both auth methods

chiedoclaude (commented):
**Honesty check (re: review 3910224274):**

Verified all claims against the actual codebase:

- **"Removed dead redirect:manual code"** — Confirmed. `handleConnect` in `OAuthCredentialRow` now uses direct navigation (line 149 of ConnectorCredentialsSection.tsx). No `fetch` or `redirect: "manual"` remains.

- **"Consolidated PROVIDER_LABELS into shared utility"** — Confirmed. `PROVIDER_LABELS` only exists in `frontend/src/lib/oauth-providers.ts`. The three consuming components (`ConnectorCredentialsSection`, `ConnectedAccountsSection`, `OAuthProviderSection`) all import from the shared file.

- **"All Go tests use t.Parallel()"** — Confirmed. 50 test functions, 53 `t.Parallel()` calls (subtests also parallelize). Each test creates its own `httptest.NewServer`, no shared mutable state.

- **"Updated docs"** — Confirmed. `docs/oauth-setup.md` has 6 references to Stripe including a full setup section. `connectors/stripe/README.md` documents both OAuth and API key auth methods.

- **"No any or unsafe casts in TypeScript"** — Confirmed. No `any` or `as` casts in the changed frontend files.

One thing I could have been clearer about in my earlier review: when I said the code "avoids leaking credentials via browser history" — that was describing the *intent* of the code I then removed. The actual fix was acknowledging that the approach didn't work and removing the dead code. The current pattern (token in query param) matches what `ConnectedAccountsSection` already uses. A proper fix would require a backend change to return the redirect URL in a JSON response instead of a 302.

chiedobot (commented):

PR #375, opened 2026-03-08
Add OAuth2 support for DocuSign connector

### LINKED ISSUE #350: Add OAuth support for DocuSign connector
## Summary

The DocuSign connector currently uses custom auth (RSA key + account info). DocuSign supports OAuth2 (both authorization code grant and JWT grant). We should add OAuth as an authentication option and make it the **default** in the UI.

## Requirements

- [ ] Add OAuth2 provider configuration for DocuSign
- [ ] OAuth should be the **default authentication method** in the UI
- [ ] **Keep existing RSA key/custom auth support** — do not remove it, add OAuth alongside it
- [ ] Users should be able to choose between OAuth and RSA key when configuring the connector
- [ ] Update frontend connector UI to present OAuth as the primary option with RSA key as an alternative

## Notes

DocuSign supports both authorization code grant (interactive) and JWT grant (service-to-service). The authorization code grant is more appropriate for the default OAuth flow.

## Context

OAuth provides a better user experience (no manual key generation, granular scope control, automatic refresh). RSA key auth remains useful for server-to-server integrations and users who prefer them.

chiedo:
- Register DocuSign as a built-in OAuth2 provider (authorization code
  grant) using account.docusign.com endpoints with the `signature` scope
- After token exchange, fetch the userinfo endpoint to capture account_id
  and base_url for the user's default account and store them in extra_data
  so the connector has everything it needs at execution time
- Update DocuSign manifest to list OAuth2 as the primary (first) credential
  and retain custom auth as an alternative, matching the OAuth-first UI
  pattern used by Slack and other multi-auth connectors
- Add TestManifest_Credentials to verify the two-credential structure
- Add fetchDocuSignUserInfo unit tests covering default account selection,
  empty-accounts error handling, and isURLExtraKey validation

Closes #350

https://claude.ai/code/session_01MBEPM4SN4LmReEv4w2kRfE


chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedobot (commented):

chiedoclaude:
## 🤖 Watch Session Summary

### Changes Made

- **DocuSign built-in OAuth provider + manifest fix** (`7049036`) — Added DocuSign as a built-in OAuth provider in `oauth/builtin.go` (AuthorizeURL, TokenURL, `signature` scope, env vars). Added `"docusign": true` to `BuiltInOAuthProviders` in `connectors/manifest.go` — without this critical map entry, manifest validation would reject the provider on startup.

- **DocuSign connector: OAuth2 as primary credential** (`b71546b`) — Updated `docusign_manifest.go` to declare OAuth2 as the first (primary/recommended) credential entry and custom/RSA key as the second (alternative), following the multi-auth pattern.

- **DocuSign post-OAuth enrichment** (`e7b33a7`) — After token exchange, the callback calls DocuSign's `/oauth/userinfo` endpoint to retrieve the user's default `account_id` and regional `base_url` (not included in token responses). Both are stored in `extra_data`. Hard-fails the callback on error (rather than storing an incomplete connection). Added SSRF protection: `base_uri` is validated to be an HTTPS `*.docusign.net` URL before storage.

- **`postOAuthEnricher` extension point** — Extracted enricher dispatch into a typed map so adding future providers requiring post-token-exchange data is a single map entry, not a handler modification.

- **Fixed Authorization header verification in tests** (`e7b33a7`) — Reworked `docuSignTestServer` helper to take the expected token and validate it server-side; previously only checked header non-empty, not the actual token value.

- **`isURLExtraKey` cleanup** — Removed `"base_url"` (dead code: `base_url` comes from stateExtraData, not tokenExtraKeys) and added clarifying comment about where DocuSign URL validation happens.

- **Merge conflict resolution** (`f531a4a`) — main branch added `calendly` provider in the same location; resolved by preserving both entries.

- **Connection pooling + missing test coverage** (`d27e0e9`) — Extracted `docuSignHTTPClient` to package level for proper TCP connection reuse. Added 5 missing tests: empty account_id, empty base_uri, HTTP (non-HTTPS) base_uri, malformed JSON response, and a registration test verifying the `"docusign"` key exists in `postOAuthEnrichers` so a typo is caught by CI.

- **Docs updates** — Added DocuSign setup section to `docs/oauth-setup.md`, updated `connectors/docusign/README.md` with OAuth authentication details, added env var comments to `.env.example`.

### Decision Log

#### ✅ Implemented
- **chiedobot**: Think about DX/UX improvements → Added OAuth2 as primary credential in manifest (recommended path first), following Slack's multi-auth pattern
- **Copilot**: `docusign` missing from `BuiltInOAuthProviders` (critical) → Fixed
- **Copilot**: Weak tests (not calling `fetchDocuSignUserInfo`) → Rewrote all tests to call the function directly via `httptest.Server`
- **Copilot**: Silent failure on userinfo fetch → Changed to hard-fail with user-visible redirect error
- **Copilot**: Dead code in `isURLExtraKey` → Removed and documented
- **chiedobot**: Fix anything a principal security engineer would balk at → Fixed dead code security comment, ensured SSRF protection validated at two layers (userinfo fetch + connector's `validateBaseURL`), verified no token/secret leakage in logs
- **chiedobot**: Refactor for DRY/maintainability → Extracted `postOAuthEnricher` type + map; kept `docuSignHTTPClient` at package level
- **chiedobot**: Improve code comments/docs → Updated README, oauth-setup.md, .env.example, added inline comments on enricher contract and SSRF protection logic
- **chiedobot**: Fix anything claimed but not done → Verified Authorization header test correctly validates token value (not just non-empty)
- **chiedobot**: Last chance — fix anything skipped as too hard → Added 5 missing edge-case tests; fixed http.Client connection pooling

#### ❌ Declined
- Full integration test for the OAuth callback handler's enricher dispatch path — requires mocking both a token exchange server AND a userinfo server simultaneously; the unit tests for `fetchDocuSignUserInfo` already cover all security-critical paths comprehensively. Adding full handler-level integration would require making `postOAuthEnrichers` injectable via `Deps`, which adds complexity for marginal gain.

#### ⚖️ Judgment Calls
- Hard-fail on userinfo fetch error (vs. soft-fail with degraded connection) — chose hard-fail because storing a connection without `account_id` would leave the user with a "Connected" status that produces confusing errors on first use; better to surface the failure immediately.
- `docuSignHTTPClient` as package-level `var` not `const` — allows tests to substitute a custom client if needed in the future; current tests use URL injection instead.

#### ❓ Open Questions
- DocuSign sandbox vs. production: the connector defaults to `https://demo.docusign.net`. OAuth users get their production `base_url` from the userinfo endpoint automatically. Custom auth users need to set `base_url` manually. Consider adding UI guidance for this.

---
*Watch session ended after 3 minutes of inactivity. Processed 7 reviews across multiple poll cycles.*

PR #444, opened 2026-03-09
Improve login rate limit UX: better error message and resend cooldown

chiedo:
- Clarify `over_email_send_rate_limit` message to tell users their
  existing OTP code is still valid if they already received one
- Add `useCooldown` hook for a 60-second countdown timer
- Show "Resend code" button on OTP step with live countdown (e.g.
  "Resend in 42s") to prevent users from hammering send and burning
  through the project-level email rate limit
- Update tests to match new message and new OtpStep props

https://claude.ai/code/session_01LrnH4zb3ErewEDcaWojVB5

### DIFF
--- a/frontend/src/auth/LoginPage.tsx
+++ b/frontend/src/auth/LoginPage.tsx
@@ -18,7 +18,11 @@ export default function LoginPage() {
         onBack={() => setStep("email")}
         onResend={async () => {
           const result = await sendOtp(email);
-          if (!result.error) cooldown.start();
+          // Start cooldown on success or rate-limit error so the button
+          // stays disabled even when the server rejects the request.
+          if (!result.error || result.error.code === "over_email_send_rate_limit") {
+            cooldown.start();
+          }
           return result;
         }}
         resendCooldownSeconds={cooldown.secondsLeft}
--- a/frontend/src/auth/LoginPage.tsx
+++ b/frontend/src/auth/LoginPage.tsx
@@ -1,5 +1,6 @@
 import { useState } from "react";
 import { useAuth } from "./AuthContext";
+import { useCooldown } from "./useCooldown";
 import EmailStep from "./EmailStep";
 import OtpStep from "./OtpStep";
 
@@ -7,6 +8,7 @@ export default function LoginPage() {
   const { sendOtp, verifyOtp } = useAuth();
   const [step, setStep] = useState<"email" | "otp">("email");
   const [email, setEmail] = useState("");
+  const cooldown = useCooldown();
 
   if (step === "otp") {
     return (
@@ -14,6 +16,12 @@ export default function LoginPage() {
         email={email}
         onVerify={(code) => verifyOtp(email, code)}
         onBack={() => setStep("email")}
+        onResend={async () => {
+          const result = await sendOtp(email);
+          if (!result.error) cooldown.start();
+          return result;
+        }}
+        resendCooldownSeconds={cooldown.secondsLeft}
       />
     );
   }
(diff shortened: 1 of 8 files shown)


chiedoclaude:
## 🤖 Watch Session Summary

### Changes Made
- **Address review comments: isResending guard, useCallback, safeErrorMessage, guard seconds<=0, add tests** (\`c645515\`) — Fixed rapid-click guard, wrapped \`start()\` in \`useCallback\`, used \`safeErrorMessage\` for resend errors, guarded \`start(0)\`, added 5 RTL tests
- **Fix pure updater violation in useCooldown and stale success message in OtpStep** (\`cc330ab\`) — Moved \`clearInterval\` out of state updater into \`useEffect\`; cleared success banner when cooldown expires
- **Start cooldown on rate-limit error** (\`f44b5cb\`) — Cooldown now starts even on \`over_email_send_rate_limit\` so button stays disabled after rejection
- **Fix screen reader announcements, tighten test queries, add cooldown expiry test** (\`79e4603\`) — Added \`aria-label\` with stable text, \`aria-hidden\` on seconds span, switched tests to \`getByRole\`, added expiry test
- **Clear resendError on cooldown expiry and add useCooldown unit tests** (\`66a83af\`) — Also clear error banner on expiry; added 6 \`useCooldown\` unit tests covering countdown, reset, start(0) no-op, unmount cleanup
- **Add role=alert/status to resend status messages** (\`90ababa\`) — Screen readers now announce resend outcomes without user navigation
- **Fix WCAG 2.5.3: align aria-label with visible text during isResending** (\`2542767\`) — \`"Resending…"\` label now matches visible text
- **Use context-specific rate-limit message on resend; shorten default** (\`54386fb\`) — Added \`overrides\` param to \`safeErrorMessage\`; OtpStep resend path shows fuller guidance, EmailStep shows shorter generic message
- **Rename shadowed error variable and guarantee spy cleanup** (\`e67796f\`) — \`resendResultError\` avoids shadowing outer \`error\`; spy restore moved to \`afterEach\`
- **Fix WCAG 2.5.3 for countdown aria-label: include seconds** (\`fa7c89b\`) — \`"Resend code in 42s (on cooldown)"\` contains the visible countdown value
- **Add test for start() default 60-second countdown** (\`22c7bee\`) — Covers the most common production code path
- **Fix override guard to use !== undefined** (\`870b6d6\`) — Prevents empty-string overrides from being silently dropped; replaced fragile \`aria-hidden\` text query with \`toHaveTextContent\`

*(None — all merges from main were clean)*

### PR Checklist Items
- ✅ **All existing tests still pass** — 661/661 throughout
- ✅ **TypeScript build has no errors** — verified before each push
- ⏭️ **Manually verify resend button in browser** — requires human action
- ⏭️ **Trigger rate limit and confirm error message** — requires human action

### Decision Log

#### ✅ Implemented
- All 20+ review comments from greptile and Copilot were addressed across multiple rounds of review
- Split \`over_email_send_rate_limit\` message: shorter generic on EmailStep, richer "if you already received a code" guidance on OtpStep resend via \`safeErrorMessage\` overrides

#### ⚖️ Judgment Calls
- WCAG 2.5.3 countdown aria-label: included the live seconds value (\`"Resend code in 42s"\`) rather than a fully static label, accepting per-second label changes as a reasonable trade-off given the button is disabled and the accessibility concern was about label mismatch, not live region noise

#### ❓ Open Questions
- None outstanding — all threads resolved

---
*Watch session ended. PR merged during session. Processed 20+ review comments across ~50 poll cycles.*

PR #469, opened 2026-03-11
Fix Google OAuth callback failing with Missing Authorization header

chiedoclaude:
## Summary
- The OAuth callback endpoint (`GET /oauth/{provider}/callback`) was wrapped in `requireSession` middleware, which requires an `Authorization: Bearer <token>` header
- When Google redirects back after user consent, it's a plain browser redirect — there is no Authorization header, so the middleware returned `{"error":{"code":"invalid_token","message":"Missing Authorization header"}}`
- Removed `requireSession` from the callback route and instead use the user identity from the signed CSRF state token (HS256, 10-min TTL), which was created during the authorize step while the user was authenticated
- The state token already encodes user ID + provider + scopes, providing equivalent authentication without needing a session header

### Claude Code
- [x] All OAuth tests pass (`go test ./api/... -run TestOAuth`)
- [x] Full backend test suite passes (`go test ./api/...`)
- [x] `go build ./api/...` compiles clean
- [x] Removed `TestOAuthCallback_UserMismatch` (no longer applicable — no session to compare)

### OpenClaw
- [ ] Open the app on mobile, tap "Connect with Google" on the connector setup modal
- [ ] Complete the Google consent screen
- [ ] Verify the callback redirects to `/settings?oauth_status=success` instead of returning a JSON error
- [ ] Verify the connection appears in the connections list

https://claude.ai/code/session_01AKyL4biF5wqtSC1yxmjQkk

### DIFF
--- a/api/oauth.go
+++ b/api/oauth.go
@@ -104,11 +104,10 @@ func init() {
 // RegisterOAuthRoutes adds OAuth-related endpoints to the mux.
 func RegisterOAuthRoutes(mux *http.ServeMux, deps *Deps) {
 	requireProfile := RequireProfile(deps)
-	requireSession := RequireSession(deps)
 
 	mux.Handle("GET /oauth/providers", requireProfile(handleListOAuthProviders(deps)))
 	mux.Handle("GET /oauth/{provider}/authorize", requireProfile(handleOAuthAuthorize(deps)))
-	mux.Handle("GET /oauth/{provider}/callback", requireSession(handleOAuthCallback(deps)))
+	mux.Handle("GET /oauth/{provider}/callback", handleOAuthCallback(deps))
 	mux.Handle("GET /oauth/connections", requireProfile(handleListOAuthConnections(deps)))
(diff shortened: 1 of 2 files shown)


chiedoclaude:
Thanks for the thorough review, Greptile. Here's how I addressed each item:

**1. Stale security comment (line 15)** — Fixed. Updated to: "Callback derives user identity from the signed state token (no session required)".

**2. Provider error before state validation** — Fixed. Moved state validation before the provider error check, so unauthenticated callers can no longer trigger frontend redirects with arbitrary `error_description` text. Added `TestOAuthCallback_ProviderErrorWithoutState` to verify this.

**3. No `jti` nonce / 10-minute replay window** — Intentionally not addressing. The provider's single-use `code` already prevents meaningful replay on the success path. For the error path, the worst case is an attacker with a leaked state token triggering repeated redirects to our own `/settings` page with a provider-controlled error message — a very low-impact scenario that doesn't justify the complexity of a nonce store (DB or Redis). The state token already expires in 10 minutes. If we later add more sensitive unauthenticated callback behavior, we can revisit.

Commit: cc33bef

PR #719, opened 2026-03-20
fix(dashboard): keep standing approval wizard open after Always Allow

chiedocursor:
## Summary
The nested **Create Standing Approval** flow (e.g. Send Email → Step 2 *Set Limits*) lives inside `ReviewApprovalDialog`. The parent dialog auto-closes after a successful approval via a timeout when `isApproved` is true and `autoCloseBlocked` is false.

There was a race: `setApproveResult` could commit before `setAutoCloseBlocked(true)`, so the auto-close timer could start and call `onOpenChange(false)` on the parent, unmounting the child wizard almost immediately.

## Changes
- `setAutoCloseBlocked(true)` is set synchronously before the `approveApproval` async call in `handleAlwaysAllow`, ensuring `autoCloseBlocked` is already `true` when any subsequent `setApproveResult` triggers a re-render.
- `autoCloseBlocked` is cleared on execution error or network failure so the normal post-approval auto-close still applies in those paths.
- `standingDialogOpen` is added as a second guard in the auto-close `useEffect`.
- `handleStandingDialogChange` always clears `autoCloseBlocked` when the nested wizard closes (the previous `!standingApprovalCreated` check was ineffective because that state updated after the close handler ran).
- `CreateStandingApprovalDialog` calls `onCreated` / `onUpdated` before `onOpenChange(false)` after a successful save.
- Regression test: parent `onOpenChange(false)` is not invoked while the wizard is visible, after waiting past `SUCCESS_AUTO_CLOSE_MS`.

### Claude Code
- [x] `npx vitest run src/pages/dashboard/__tests__/ReviewApprovalDialog.test.tsx`
- [x] `npx vitest run src/pages/dashboard/__tests__/CreateStandingApprovalDialog.test.tsx`

### Manual Testing
- [ ] From a pending approval with parameters, choose **Always allow this action**; confirm the standing-approval wizard stays open through Step 1 (constraints) and Step 2 (limits) without the parent dialog disappearing.
- [ ] After saving limits, confirm success flow still behaves as expected (nested closes, parent shows approved state / Done).

<div><a href="https://cursor.com/agents/bc-2ef04fe6-cf56-433a-8894-89d664f95492"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2ef04fe6-cf56-433a-8894-89d664f95492"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/frontend/src/pages/dashboard/CreateStandingApprovalDialog.tsx
+++ b/frontend/src/pages/dashboard/CreateStandingApprovalDialog.tsx
@@ -417,8 +417,8 @@ export function CreateStandingApprovalDialog({
         });
         toast.success("Standing approval updated");
         resetForm();
-        onOpenChange(false);
         onUpdated?.();
+        onOpenChange(false);
       } else {
         await createStandingApproval({
           agent_id: agentId,
@@ -431,8 +431,8 @@ export function CreateStandingApprovalDialog({
         });
         toast.success("Standing approval created");
         resetForm();
-        onOpenChange(false);
         onCreated?.();
+        onOpenChange(false);
       }
     } catch (err) {
       toast.error(
--- a/frontend/src/pages/dashboard/ReviewApprovalDialog.tsx
+++ b/frontend/src/pages/dashboard/ReviewApprovalDialog.tsx
@@ -196,7 +196,10 @@ export function ReviewApprovalDialog({
 
   function handleStandingDialogChange(nextOpen: boolean) {
     setStandingDialogOpen(nextOpen);
-    if (!nextOpen && !standingApprovalCreated) {
+    // Unblock parent auto-close whenever the nested wizard closes (cancel or success).
+    // Do not gate on standingApprovalCreated — that state updates after this handler runs,
+    // so a closure check would always see the pre-create value.
+    if (!nextOpen) {
       setAutoCloseBlocked(false);
     }
   }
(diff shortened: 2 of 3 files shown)


chiedocursor:
Addressed the Greptile feedback in 8e12cfcd:

1. **Regression test** — Added coverage that `onOpenChange(false)` is not called while the standing-approval wizard is open, waiting **3.1s** (past `SUCCESS_AUTO_CLOSE_MS`). Used real `setTimeout` instead of `vi.useFakeTimers()` for the whole test because fake timers caused `waitFor` to hang and could leak into the next test on timeout.

2. **`handleStandingDialogChange`** — Removed the `!standingApprovalCreated` branch; it always saw a stale `false`. We always need `setAutoCloseBlocked(false)` when the nested wizard closes (cancel or success) so the parent auto-close can run. Comment documents why.

3. **`CreateStandingApprovalDialog`** — `onCreated` / `onUpdated` now run before `onOpenChange(false)` so the parent is notified before the wizard closes (clearer ordering; React state from `onCreated` still applies on the next render).

PR #763, opened 2026-03-22
Fix credentials_ready to honor agent_connector_credentials bindings

### LINKED ISSUE #760: credentials_ready query ignores agent_connector_credentials bindings
## Problem

The `credentials_ready` check in `db/capabilities.go` is stale — it still queries the global `credentials` and `oauth_connections` tables directly instead of using the `agent_connector_credentials` binding table introduced in migration `20260313104413`.

This causes agents to report `credentials_ready: false` even when connectors are properly registered and have credentials bound.

### Root cause

The capabilities query (lines 75–92) checks:

```sql
NOT EXISTS (
    SELECT 1 FROM connector_required_credentials crc
    WHERE crc.connector_id = c.id
      AND NOT EXISTS (
        SELECT 1 FROM credentials cr
        WHERE cr.user_id = ac.approver_id
          AND cr.service = crc.service        -- checks connector_required_credentials.service
      )
      AND NOT EXISTS (
        SELECT 1 FROM oauth_connections oc
        WHERE oc.user_id = ac.approver_id
          AND oc.provider = crc.oauth_provider
          AND oc.status = 'active'
          AND crc.oauth_scopes <@ oc.scopes
      )
) AS credentials_ready
```

Two bugs:

1. **Service name mismatch**: The credential assignment endpoint (`api/agent_connector_credentials.go:67`) validates `credentials.service == connector_id`, but the capabilities query checks `credentials.service == connector_required_credentials.service`. If those differ (e.g., connector ID `"google-calendar"` with required credential service `"google"`), `credentials_ready` is always false.

2. **Ignores bindings entirely**: Even when service names happen to match, the query checks whether the user has *any* credential for that service globally — not whether a credential is actually *bound* to this specific agent+connector pair via `agent_connector_credentials`. This was the whole point of the binding table.

### Affected files

- `db/capabilities.go` — the stale query (lines 75–92)
- `db/capabilities_test.go` — tests pass because fixtures use matching service names and don't test the binding path

### Developer
- [ ] Update the `credentials_ready` SQL to check `agent_connector_credentials` instead of doing direct lookups against `credentials`/`oauth_connections`
- [ ] A connector should be `credentials_ready = true` when either: it has no rows in `connector_required_credentials`, OR it has a binding row in `agent_connector_credentials` for this agent+connector pair
- [ ] Update tests to cover the binding-based check (including the case where a global credential exists but is not bound)
- [ ] Verify the fix works end-to-end with a connector whose `required_credentials.service` differs from the connector ID

### Manual Testing
- [ ] Confirm an agent with bound credentials now reports `credentials_ready: true` in the capabilities response

chiedocursor:
## Summary

Fixes [issue #760](https://github.com/supersuit-tech/permission-slip/issues/760): `credentials_ready` in capabilities no longer relied on global `credentials` / `oauth_connections` lookups alone. It now requires a row in `agent_connector_credentials` for the agent+connector, with the bound credential validated against `connector_required_credentials` (static: `credentials.service` equals the required `service` **or** the connector id, aligned with assign-credential validation; OAuth: provider + scopes on the **bound** connection).

Documentation and OpenAPI descriptions were updated so agents understand that “stored on the account” is not enough without an assignment.

## Test plan

- `make build` (passes locally)
- `go test -c ./api/... ./db/...` (packages compile)
- **Not run here:** `make test-backend` / DB tests — PostgreSQL was unavailable in this environment; CI should run the full suite.

### Developer

- [ ] Confirm CI green (`make test-backend`, integration as applicable)

### Manual Testing

- [ ] Enable a connector for an agent, assign credentials via the binding flow, call `GET /agents/{id}/capabilities` and confirm `credentials_ready: true`
- [ ] With the same user credential stored but **not** assigned to the agent, confirm `credentials_ready: false` and `credentials_setup_url` present

<div><a href="https://cursor.com/agents/bc-7c5fbaf3-9406-4bd3-a3cf-b8e6ec970782"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7c5fbaf3-9406-4bd3-a3cf-b8e6ec970782"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/api/capabilities_test.go
+++ b/api/capabilities_test.go
@@ -35,7 +35,8 @@ func TestGetCapabilities_HappyPath(t *testing.T) {
 	testhelper.InsertConnectorWithDescription(t, tx, conn1, "Gmail", "Send and manage emails")
 	svc1 := testhelper.GenerateID(t, "svc_")
 	testhelper.InsertConnectorRequiredCredential(t, tx, conn1, svc1, "api_key")
-	testhelper.InsertCredential(t, tx, testhelper.GenerateID(t, "cred_"), uid, svc1)
+	cred1 := testhelper.GenerateID(t, "cred_")
+	testhelper.InsertCredential(t, tx, cred1, uid, svc1)
 
 	riskLow := "low"
 	sendDesc := "Send an email"
@@ -46,6 +47,7 @@ func TestGetCapabilities_HappyPath(t *testing.T) {
 		ParametersSchema: schema,
 	})
 	testhelper.InsertAgentConnector(t, tx, agentID, uid, conn1)
+	testhelper.InsertAgentConnectorCredential(t, tx, testhelper.GenerateID(t, "acc_"), agentID, uid, conn1, cred1)
 
 	// Add a standing approval.
 	saID := testhelper.GenerateID(t, "sa_")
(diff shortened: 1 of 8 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---
*Watch session ended after 6 minutes of inactivity. Processed 0 comments across 6 poll cycles.*

PR #936, opened 2026-04-13
Add optional gateway secret for private deployments

chiedoclaude:
## Summary

- **Backend:** Adds `GatewaySecretMiddleware` — when the `GATEWAY_SECRET` env var is set, all non-OPTIONS requests must include a matching `X-Gateway-Secret` header or receive 403. Runs as the outermost middleware (before CORS, security headers, routing, auth) using constant-time comparison. When unset, the middleware is a complete no-op.
- **Mobile:** Adds a "Custom Server" section in Settings where users can configure a custom host URL and gateway secret. The API client middleware rewrites request URLs and injects the header on every request. Config is persisted in SecureStore (iOS Keychain / Android EncryptedSharedPreferences).
- **CORS:** Adds `X-Gateway-Secret` to the allowed CORS headers so browser-based clients can send the header in cross-origin requests.

### Developer
- [x] Backend: 8 unit tests for gateway secret middleware (no secret passthrough, valid/invalid/missing/empty header, OPTIONS exempt, all HTTP methods, timing safety)
- [x] Mobile: All 225 existing tests pass with new SecureStore mock
- [x] TypeScript check passes (pre-existing errors only)

### Manual Testing
- [ ] Deploy with `GATEWAY_SECRET=<random-64-hex>` and verify requests without the header get 403
- [ ] Deploy without `GATEWAY_SECRET` and verify everything works as before (no-op)
- [ ] In mobile app Settings, enable Custom Server, enter host URL + gateway secret, save, restart app, verify API calls route to the custom host with the header
- [ ] Toggle Custom Server off, restart, verify app reverts to default host
- [ ] Verify push notifications still work when connected to a custom host

https://claude.ai/code/session_017gCSx2YrGjQVXEy5zVHQT3

### DIFF
--- a/api/gateway_secret_middleware.go
+++ b/api/gateway_secret_middleware.go
@@ -1,0 +1,51 @@
+package api
+
+import (
+	"crypto/subtle"
+	"log"
+	"net/http"
+)
+
+// GatewaySecretMiddleware returns middleware that rejects requests without a
+// valid X-Gateway-Secret header. When secret is empty, the middleware is a
+// no-op — all requests pass through unchanged.
+//
+// This is designed for private deployments behind Cloudflare Tunnel: set the
+// GATEWAY_SECRET env var to a long random string and configure clients (mobile
+// app, curl, etc.) to send the same value in the X-Gateway-Secret header.
+//
+// The check runs as the outermost middleware so unauthorized requests are
+// rejected with minimal processing (before CORS, security headers, routing,
+// auth, etc.). CORS preflight (OPTIONS) requests are exempt because browsers
+// send them without custom headers.
+func GatewaySecretMiddleware(secret string) func(http.Handler) http.Handler {
+	if secret != "" {
+		log.Println("Gateway secret: enabled — requests without a valid X-Gateway-Secret header will be rejected")
+	}
+
+	return func(next http.Handler) http.Handler {
+		if secret == "" {
+			return next
+		}
+
+		secretBytes := []byte(secret)
+
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Allow CORS preflight through — browsers send OPTIONS without
+			// custom headers, so blocking them would break cross-origin
+			// requests entirely.
+			if r.Method == http.MethodOptions {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			provided := r.Header.Get("X-Gateway-Secret")
+			if subtle.ConstantTimeCompare([]byte(provided), secretBytes) != 1 {
+				http.Error(w, "Forbidden", http.StatusForbidden)
+				return
+			}
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
--- a/api/gateway_secret_middleware.go
+++ b/api/gateway_secret_middleware.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"crypto/sha256"
 	"crypto/subtle"
 	"log"
 	"net/http"
@@ -14,10 +15,17 @@ import (
 // GATEWAY_SECRET env var to a long random string and configure clients (mobile
 // app, curl, etc.) to send the same value in the X-Gateway-Secret header.
 //
-// The check runs as the outermost middleware so unauthorized requests are
-// rejected with minimal processing (before CORS, security headers, routing,
-// auth, etc.). CORS preflight (OPTIONS) requests are exempt because browsers
-// send them without custom headers.
+// Middleware ordering: in main.go this runs inside SecurityHeadersMiddleware
+// (which only sets response headers and never blocks), but outside CORS,
+// routing, and auth — so unauthorized requests are rejected before any
+// application processing. True CORS preflight requests (OPTIONS with an
+// Access-Control-Request-Method header) are exempt because browsers send them
+// without custom headers; other OPTIONS requests are still gated.
+//
+// The header and configured secret are SHA-256 hashed before comparison so
+// that the fixed-length digests (always 32 bytes) feed ConstantTimeCompare.
+// This avoids leaking the secret length via an early length-mismatch return
+// inside ConstantTimeCompare.
 func GatewaySecretMiddleware(secret string) func(http.Handler) http.Handler {
 	if secret != "" {
 		log.Println("Gateway secret: enabled — requests without a valid X-Gateway-Secret header will be rejected")
@@ -28,19 +36,19 @@ func GatewaySecretMiddleware(secret string) func(http.Handler) http.Handler {
 			return next
 		}
 
-		secretBytes := []byte(secret)
+		secretSum := sha256.Sum256([]byte(secret))
 
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			// Allow CORS preflight through — browsers send OPTIONS without
-			// custom headers, so blocking them would break cross-origin
-			// requests entirely.
-			if r.Method == http.MethodOptions {
+			// Exempt only genuine CORS preflights — OPTIONS requests that
+			// include Access-Control-Request-Method. Bare OPTIONS requests
+			// still require the gateway secret so nothing slips through.
+			if r.Method == http.MethodOptions && r.Header.Get("Access-Control-Request-Method") != "" {
 				next.ServeHTTP(w, r)
 				return
 			}
 
-			provided := r.Header.Get("X-Gateway-Secret")
-			if subtle.ConstantTimeCompare([]byte(provided), secretBytes) != 1 {
+			providedSum := sha256.Sum256([]byte(r.Header.Get("X-Gateway-Secret")))
+			if subtle.ConstantTimeCompare(providedSum[:], secretSum[:]) != 1 {
 				http.Error(w, "Forbidden", http.StatusForbidden)
 				return
 			}
--- a/main.go
+++ b/main.go
@@ -441,6 +441,12 @@ func main() {
 	// handler, including the health check and SPA handler.
 	handler := api.CORSMiddleware(deps.AllowedOrigins)(mux)
 
+	// Gateway secret — optional access gate for private deployments.
+	// When GATEWAY_SECRET is set, every non-OPTIONS request must include a
+	// matching X-Gateway-Secret header or receive 403. This is the outermost
+	// middleware so unauthorized requests are rejected before any processing.
+	handler = api.GatewaySecretMiddleware(os.Getenv("GATEWAY_SECRET"))(handler)
+
 	// Wrap all routes with security headers (outermost layer).
 	// Include the Supabase URL in CSP connect-src so the frontend can reach
 	// the auth/API endpoints in production. Sentry's ingest domain is always
--- a/main.go
+++ b/main.go
@@ -442,9 +442,12 @@ func main() {
 	handler := api.CORSMiddleware(deps.AllowedOrigins)(mux)
 
 	// Gateway secret — optional access gate for private deployments.
-	// When GATEWAY_SECRET is set, every non-OPTIONS request must include a
-	// matching X-Gateway-Secret header or receive 403. This is the outermost
-	// middleware so unauthorized requests are rejected before any processing.
+	// When GATEWAY_SECRET is set, every request (except genuine CORS
+	// preflights) must include a matching X-Gateway-Secret header or receive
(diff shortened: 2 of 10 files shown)


api/gateway_secret_middleware.go
```diff
@@ -0,0 +1,51 @@
...
+			// Allow CORS preflight through — browsers send OPTIONS without
+			// custom headers, so blocking them would break cross-origin
+			// requests entirely.
+			if r.Method == http.MethodOptions {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			provided := r.Header.Get("X-Gateway-Secret")
+			if subtle.ConstantTimeCompare([]byte(provided), secretBytes) != 1 {
```
chiedocursor: **[High · Security]** Using `subtle.ConstantTimeCompare` on the raw header bytes compares slices of potentially different lengths. In Go, `ConstantTimeCompare` returns immediately when lengths differ, which can leak timing information about the configured secret length and weakens the stated goal of a timing-safe check.

**Why:** An attacker who can observe response timings may iterate on candidate lengths before mounting an offline or online search against the secret.

**Fix:** Normalize both sides to a fixed-length digest before comparing, e.g. hash header and secret with SHA-256 and compare the 32-byte outputs:

```go
providedSum := sha256.Sum256([]byte(provided))
secretSum := sha256.Sum256(secretBytes)
if subtle.ConstantTimeCompare(providedSum[:], secretSum[:]) != 1 {
    http.Error(w, "Forbidden", http.StatusForbidden)
    return
}
```

(Use `crypto/sha256`.)

mobile/App.tsx
```diff
@@ -139,6 +140,12 @@ function AppContent({ onRetry }: { onRetry: () => void }) {
 }
 
 export default function App() {
+  // Hydrate custom host config from SecureStore so the API client middleware
+  // can route requests to a private deployment if configured.
+  useEffect(() => {
```
chiedocursor, line 148: **[High · Architecture]** `loadCustomHostConfig()` is fired in `App` without awaiting completion, while the API `client` module is imported at startup and child components can fire requests on first mount.

**Why:** The first one or more API calls after cold start can still use the default production host and omit `X-Gateway-Secret`, causing surprising 403s against a gateway-locked private server even though settings were saved previously.

**Fix:** Gate rendering of authenticated/networked subtrees on `await loadCustomHostConfig()` (e.g. show splash until hydration completes), or hydrate before `createClient`/`client.use` runs (init module that awaits before exporting client).

main.go
```diff
@@ -441,6 +441,12 @@ func main() {
 	// handler, including the health check and SPA handler.
 	handler := api.CORSMiddleware(deps.AllowedOrigins)(mux)
 
+	// Gateway secret — optional access gate for private deployments.
```
chiedocursor, line 457: **[High · Documentation / Correctness]** Comments claim this is the "outermost" middleware before CORS, but the actual HTTP chain is `SecurityHeaders → GatewaySecret → CORSMiddleware → mux` (`SecurityHeadersMiddleware` is applied after this line).

**Why:** Operators and future maintainers may place the secret in the wrong trust boundary (e.g. expecting zero security-header processing for anonymous probes) and the description in the PR diverges from runtime behavior.

**Fix:** Update comments (PR body + inline) to describe the real ordering, or reorder wrappers so gateway truly sits outside security headers if that was the intent.

api/gateway_secret_middleware.go
```diff
@@ -0,0 +1,51 @@
...
+			return next
+		}
+
+		secretBytes := []byte(secret)
+
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Allow CORS preflight through — browsers send OPTIONS without
+			// custom headers, so blocking them would break cross-origin
+			// requests entirely.
+			if r.Method == http.MethodOptions {
```
chiedocursor: **[Medium · Security]** All `OPTIONS` requests skip the gateway check, not just CORS preflight (`Access-Control-Request-Method` present).

**Why:** Any endpoint that performs non-trivial work on `OPTIONS` (unusual but possible with proxies) would bypass the gate.

**Fix:** If you only need browser preflights, mirror `CORSMiddleware` and exempt only `OPTIONS` requests that include `Access-Control-Request-Method` (and optionally `Access-Control-Request-Headers` mentioning the gateway header).

chiedocursor (commented):
## Review Round 1/6

**Findings:** 4 (0 critical, 3 high, 1 medium)

The gateway/feature wiring is clear and well-tested on the backend, but there are a few concrete issues worth fixing before merge: Go `ConstantTimeCompare` leaks when lengths differ; the mobile client can issue requests before SecureStore hydration finishes; and a few comments overstate the middleware ordering relative to security headers.

### What Looks Good
- Sensible split between SecureStore persistence and a synchronous in-memory cache for the fetch middleware.
- Broad unit coverage for the gateway middleware (methods, OPTIONS exemption, invalid/missing header, no-op when unset).
- CORS preflight explicitly allows `X-Gateway-Secret`, which unblocks browser clients that need the header.

---

api/gateway_secret_middleware.go
chiedoclaude: Fixed in 45a551e — both sides are now SHA-256 hashed before `subtle.ConstantTimeCompare`, so the comparison runs on fixed 32-byte digests regardless of secret length.

mobile/App.tsx
chiedoclaude: Fixed in 45a551e — `App` now gates rendering on a `hostHydrated` state that flips to true only after `loadCustomHostConfig()` resolves. Until then, a splash with an ActivityIndicator is shown, so no API call can fire before the cache is populated.

main.go
chiedoclaude: Fixed in 45a551e — updated the comments in both `main.go` and `gateway_secret_middleware.go` to describe the real chain: `SecurityHeaders → Gateway → CORS → mux`. I kept the current order because `SecurityHeadersMiddleware` only sets response headers (never blocks), so wrapping Gateway inside it still gives defense-in-depth headers on 403 responses without weakening the gate's trust boundary.

api/gateway_secret_middleware.go
chiedoclaude: Fixed in 45a551e — the OPTIONS exemption now requires `Access-Control-Request-Method` to be set (matching how `CORSMiddleware` detects real preflights). Bare OPTIONS requests are gated like any other method. Added a `BareOPTIONS_Rejects` test covering the new behavior.

mobile/src/lib/customHostConfig.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * In-memory cache so middleware reads are synchronous and fast.
+ * Call loadCustomHostConfig() at app startup to hydrate from SecureStore.
+ */
+let cachedHost: string | null = null;
+let cachedSecret: string | null = null;
+
+/**
+ * Hydrate the in-memory cache from SecureStore. Call once at app startup
+ * before creating the API client.
```
chiedocursor: **[Low · Documentation]** The docstring still says to call `loadCustomHostConfig()` "before creating the API client", but `mobile/src/api/client.ts` constructs `createClient(...)` at module import time — before `App` runs — while hydration is now enforced at the UI layer.

**Why:** The next person touching startup may follow the comment, remove the `App` gate, and reintroduce the race without realizing it.

**Fix:** Reword to something like: "Hydrate from SecureStore during app startup (see `App.tsx`); middleware reads the in-memory cache synchronously."

chiedocursor (commented):
## Review Round 2/6

**Findings:** 1 (0 critical, 0 high, 0 medium, 1 low)

### Changes Since Last Round
- `45a551e` — **Address review:** SHA-256 digests before `ConstantTimeCompare`, stricter CORS preflight exemption for `OPTIONS`, and an app-level hydration gate so API traffic waits for `loadCustomHostConfig()`.
- `50b9915` — merge of `origin/main` (no additional files in the PR diff).

### Remaining Issues
- **Low:** `customHostConfig` module docs still describe hydration happening before the API client is constructed, which is no longer true now that `client.ts` is imported at module load.

### What Looks Good
- The gateway middleware changes directly address the length-leak and bare-`OPTIONS` bypass concerns, with matching test coverage (`CORSPreflightExempt` / `BareOPTIONS_Rejects`).
- `App.tsx` now blocks the navigator until host settings hydrate, which closes the early-request race cleanly.
- `main.go` comments now accurately describe ordering relative to `SecurityHeadersMiddleware`.

---

mobile/src/lib/customHostConfig.ts
chiedoclaude: Fixed in 69f59e1 — reworded the docstring to describe the real startup sequence: hydration happens during `App` startup (which gates rendering), while the middleware reads the in-memory cache synchronously per request.

chiedocursor (commented):
## Review Round 3/6

**Findings:** 0 (0 critical, 0 high, 0 medium, 0 low)

### Changes Since Last Round
- `69f59e1` — **Fix stale docstring in customHostConfig.ts:** documentation now matches the actual startup order (`client.ts` import vs `App.tsx` hydration gate).

### What Looks Good
- Quick follow-up on the remaining documentation nit; the module-level comment now explains why the UI gate matters, which should prevent accidental regressions.

---
*Automated review by `/review` — Round 3 of 6*

chiedocursor (commented):
## Review Round 4/6

**Findings:** 0 (0 critical, 0 high, 0 medium, 0 low)

No new commits since last round — re-examination of the gateway + mobile wiring at `69f59e1` did not surface additional issues beyond what was already fixed.

### What Looks Good
- Stable head with the docstring correction merged; the custom-host story is internally consistent from SecureStore → cache → fetch middleware → UI gate.

---
*Automated review by `/review` — Round 4 of 6*

chiedocursor (commented):
## Review Round 5/6

**Findings:** 0 (0 critical, 0 high, 0 medium, 0 low)

No new commits since last round — no additional issues found on re-examination.

### What Looks Good
- PR head remains stable; nothing new to flag in the gateway secret + custom host path.

---
*Automated review by `/review` — Round 5 of 6*

chiedocursor (commented):
## Review Round 6/6

**Findings:** 0 (0 critical, 0 high, 0 medium, 0 low)

### Changes Since Last Round
- `85fa81d` — merge of `origin/main` into the PR branch (no net changes to the gateway/custom-host files in the aggregate diff beyond staying current with `main`).

Re-checked the gateway middleware, CORS allow-list, `main.go` wiring, and the mobile hydration gate against the new head — no regressions spotted.

### What Looks Good
- Keeping the branch rebased/merged with `main` reduces integration risk while the feature set itself remains sound.

---
*Automated review by `/review` — Round 6 of 6*

chiedocursor:
## Code Review Complete

**PR:** #936 — Add optional gateway secret for private deployments
**Author:** @chiedoclaude
**Rounds completed:** 6/6
**Reason for stopping:** completed all rounds

### Outstanding Issues
None — prior findings were addressed in follow-up commits (`45a551e` for SHA-256 comparison, stricter CORS preflight handling, and the mobile hydration gate; `69f59e1` for the `customHostConfig` docstring). Latest head `85fa81d` is a merge with `origin/main` that did not reintroduce concerns in the reviewed surface area.

### Review History
|-------|------------------|-------|
| 1 | 4 | Flagged `ConstantTimeCompare` length leak, early mobile hydration race, middleware ordering docs vs reality, and blanket `OPTIONS` bypass |
| 2 | 1 | Remaining stale docstring in `customHostConfig.ts` after fixes landed |
| 3 | 0 | Docstring follow-up merged (`69f59e1`) |
| 4 | 0 | Stable head — no new findings |
| 5 | 0 | Stable head — no new findings |
| 6 | 0 | Post-merge-with-main check — no regressions spotted |

### Overall Assessment
The feature is in good shape for merge from a security/architecture perspective after the follow-ups: the gateway check now compares fixed-length digests, only genuine browser preflights bypass the secret, mobile startup blocks UI until SecureStore hydration completes (avoiding accidental hits to the default host), and documentation matches the real middleware order relative to `SecurityHeadersMiddleware`. The original review’s inline notes on `gateway_secret_middleware.go` landed without diff line anchors (GitHub reported `line: null`), but the substance was incorporated into the code.

Strong positives include thorough unit tests for the middleware edge cases and a clear separation between SecureStore persistence and synchronous middleware reads.

---

Issue #962, opened 2026-04-20
Support multiple instances of the same connector type per agent


chiedoclaude:
## Context

Today an agent can have at most one Slack, one Google, one GitHub, etc. — the constraint is enforced by `UNIQUE (agent_id, connector_id)` on `agent_connector_credentials` (migration `20260313104413`) and by the composite PK on `agent_connectors`. Users want to attach N instances of the same connector type to a single agent (e.g. 5 Slack workspaces).

We've already lifted the one-per-provider limit on `oauth_connections` (migration `20260313201757`) — this issue extends the same shape to agent-connector bindings end-to-end.

## Recommended approach

Add a surrogate `connector_instance_id UUID` + human `label` on `agent_connectors`. Backfill one default instance per existing (agent, connector) pair so Day 1 matches Day 0. The action type string (`slack.send_message`) does **not** change. Agents pick an instance by passing an optional `connector_instance` parameter (label or UUID); the server resolves it and freezes the UUID + label into the approval's `action` JSON so approvers (web, mobile, future CLI) see which workspace is targeted. New `/instances` sub-routes coexist with the existing `/credential` routes, which are repointed to the default instance for BC.

### Key decisions

- **Instance selection is a parameter, not a new tool name.** Preserves stable `action_type` strings that standing approvals, `action_configurations`, and `Registry.GetAction`'s dot-split all depend on. The capabilities endpoint lists each agent's instance labels in the tool description so the LLM picks by label naturally.
- **Omission rules:** 0 instances → existing "no credential assigned" error; 1 instance → auto-select (BC); 2+ → 400 `connector_instance_required` with `available_instances: [labels...]`.
- **Approval payload:** `_connector_instance_id` + `_connector_instance_label` are stashed at the top level of the existing `action` JSONB. No `approvals` schema change.
- **Standing approvals:** nullable `connector_instance_id` column. NULL = type-wide (legacy), non-null = instance-scoped.
- **No feature flag.** One deploy is safe: backfill ensures old code + new schema both work.

## Data model sketch

**`agent_connectors`**
- Add `connector_instance_id uuid NOT NULL DEFAULT gen_random_uuid()`
- Add `label text NOT NULL` with `UNIQUE (agent_id, connector_id, label)`
- Add `is_default bool NOT NULL DEFAULT false` with partial unique `UNIQUE (agent_id, connector_id) WHERE is_default`
- New PK: `(agent_id, approver_id, connector_id, connector_instance_id)`

[...]

PR #963, opened 2026-04-20
feat(db): connector instance schema + backfill (issue #962 phase 1)

chiedocursor:
## Summary

Implements **Phase 1** of #962: schema migrations and backfill for multiple connector instances per agent.

- **`agent_connectors`**: `connector_instance_id`, `label`, `is_default`; new PK `(agent_id, approver_id, connector_id, connector_instance_id)`; unique `(agent_id, connector_id, label)`; partial unique default per `(agent_id, connector_id)`. Backfill sets `label` from `connectors.name` and marks existing rows default.
- **`agent_connector_credentials`**: `connector_instance_id` with FK to the new PK; unique `(agent_id, connector_id, connector_instance_id)`; backfill from matching `agent_connectors` rows.
- **`standing_approvals`**: nullable `connector_instance_id` + index `(agent_id, action_type, status, connector_instance_id)`.
- **Triggers** keep legacy `INSERT` shapes working: `agent_connectors` fills `label` / first default; `agent_connector_credentials` resolves omitted `connector_instance_id` to the default instance.
- **`supabase/seed.sql`**: minimal valid seed so `supabase db reset` (configured to load `./seed.sql`) succeeds after migrations.

Application updates (minimal, required for the new constraints): `EnableAgentConnector` / `UpsertAgentConnectorCredential` conflict targets, `DisableAgentConnector` scalar subqueries + extra `RETURNING` column, schema tests.

### Developer

- [ ] `make test-backend` (or CI) — includes `TestMigrationTimestampsUnique`, RLS migration tests, and updated `db` schema tests.
- [ ] Apply migrations on a dev DB with seed data; confirm one default instance per former `(agent, connector)` row and credential rows keyed by that instance.

### Manual Testing

- [ ] `make migrate-up` / `supabase db reset` — migrations apply cleanly.
- [ ] Spot-check: `agent_connectors.label` matches `connectors.name` for seeded rows; `is_default = true` for those rows.

Refs #962 (Phase 1 only — does not complete the full issue).

<div><a href="https://cursor.com/agents/bc-bb19f854-ed5b-4115-9ca7-53bbd6f2afe6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bb19f854-ed5b-4115-9ca7-53bbd6f2afe6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/db/agent_connector_credentials_test.go
+++ b/db/agent_connector_credentials_test.go
@@ -1,0 +1,17 @@
+package db_test
+
+import (
+	"testing"
+
+	"github.com/supersuit-tech/permission-slip/db/testhelper"
+)
+
+func TestAgentConnectorCredentialsSchema(t *testing.T) {
+	t.Parallel()
+	tx := testhelper.SetupTestDB(t)
+
+	testhelper.RequireColumns(t, tx, "agent_connector_credentials", []string{
+		"id", "agent_id", "connector_id", "approver_id", "connector_instance_id",
+		"credential_id", "oauth_connection_id", "created_at",
+	})
+}
--- a/db/agent_connectors.go
+++ b/db/agent_connectors.go
@@ -99,7 +99,7 @@ func EnableAgentConnector(ctx context.Context, db DBTX, agentID int64, approverI
 			INSERT INTO agent_connectors (agent_id, approver_id, connector_id)
 			SELECT $1, $2, $3
 			WHERE EXISTS (SELECT 1 FROM agent_check)
-			ON CONFLICT (agent_id, connector_id) DO NOTHING
+			ON CONFLICT (agent_id, connector_id, label) DO NOTHING
 			RETURNING agent_id, connector_id, enabled_at
 		)
 		SELECT agent_id, connector_id, enabled_at FROM inserted
@@ -140,7 +140,7 @@ func DisableAgentConnector(ctx context.Context, db DBTX, agentID int64, approver
 		WITH deleted AS (
 			DELETE FROM agent_connectors
 			WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
-			RETURNING agent_id, connector_id
+			RETURNING agent_id, connector_id, connector_instance_id
 		), revoked AS (
 			UPDATE standing_approvals
 			SET status = 'revoked', revoked_at = now()
@@ -155,8 +155,8 @@ func DisableAgentConnector(ctx context.Context, db DBTX, agentID int64, approver
 			RETURNING 1
 		)
 		SELECT
-			(SELECT agent_id FROM deleted),
-			(SELECT connector_id FROM deleted),
+			(SELECT agent_id FROM deleted LIMIT 1),
+			(SELECT connector_id FROM deleted LIMIT 1),
 			now(),
 			(SELECT count(*) FROM revoked)`,
 		agentID, approverID, connectorID,
--- a/db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
+++ b/db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
@@ -1,0 +1,78 @@
+-- +goose Up
+-- Multi-instance connector support: surrogate instance id + label + default flag per (agent, connector).
+-- Backfill: one instance per existing row; label from connectors.name; all rows are default.
+
+ALTER TABLE agent_connectors
+    ADD COLUMN connector_instance_id uuid NOT NULL DEFAULT gen_random_uuid(),
+    ADD COLUMN label text,
+    ADD COLUMN is_default boolean NOT NULL DEFAULT false;
+
+UPDATE agent_connectors ac
+SET label = c.name
+FROM connectors c
+WHERE c.id = ac.connector_id;
+
+ALTER TABLE agent_connectors
+    ALTER COLUMN label SET NOT NULL;
+
+-- Exactly one default instance per (agent_id, connector_id) — every existing row is that default.
+UPDATE agent_connectors SET is_default = true;
+
+CREATE UNIQUE INDEX uq_agent_connectors_agent_connector_label
+    ON agent_connectors (agent_id, connector_id, label);
+
+CREATE UNIQUE INDEX uq_agent_connectors_default_per_pair
+    ON agent_connectors (agent_id, connector_id)
+    WHERE is_default;
+
+-- Back-compat: allow INSERT with only (agent_id, approver_id, connector_id); fill label and default flag.
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION trg_agent_connectors_before_insert()
+RETURNS trigger AS $$
+BEGIN
+    IF NEW.label IS NULL THEN
+        SELECT name INTO STRICT NEW.label FROM connectors WHERE id = NEW.connector_id;
+    END IF;
+    -- First row for (agent, connector) must be the default instance (column default is false).
+    IF NOT EXISTS (
+        SELECT 1 FROM agent_connectors
+        WHERE agent_id = NEW.agent_id
+          AND connector_id = NEW.connector_id
+          AND is_default
+    ) THEN
+        NEW.is_default := true;
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+-- +goose StatementEnd
+
+CREATE TRIGGER trg_agent_connectors_before_insert
+    BEFORE INSERT ON agent_connectors
+    FOR EACH ROW
+    EXECUTE FUNCTION trg_agent_connectors_before_insert();
+
+ALTER TABLE agent_connectors DROP CONSTRAINT agent_connectors_pkey;
+
+ALTER TABLE agent_connectors
+    ADD CONSTRAINT agent_connectors_pkey
+    PRIMARY KEY (agent_id, approver_id, connector_id, connector_instance_id);
+
+-- +goose Down
+
+DROP TRIGGER IF EXISTS trg_agent_connectors_before_insert ON agent_connectors;
+DROP FUNCTION IF EXISTS trg_agent_connectors_before_insert();
+
+ALTER TABLE agent_connectors DROP CONSTRAINT agent_connectors_pkey;
+
+ALTER TABLE agent_connectors
+    ADD CONSTRAINT agent_connectors_pkey
+    PRIMARY KEY (agent_id, connector_id);
+
+DROP INDEX IF EXISTS uq_agent_connectors_default_per_pair;
+DROP INDEX IF EXISTS uq_agent_connectors_agent_connector_label;
+
+ALTER TABLE agent_connectors
+    DROP COLUMN connector_instance_id,
+    DROP COLUMN label,
+    DROP COLUMN is_default;
--- a/db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
+++ b/db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
@@ -31,7 +31,12 @@ CREATE OR REPLACE FUNCTION trg_agent_connectors_before_insert()
 RETURNS trigger AS $$
 BEGIN
     IF NEW.label IS NULL THEN
-        SELECT name INTO STRICT NEW.label FROM connectors WHERE id = NEW.connector_id;
+        SELECT c.name INTO NEW.label FROM connectors c WHERE c.id = NEW.connector_id;
+        IF NEW.label IS NULL THEN
+            -- Match FK violation so API maps to connector_not_found (STRICT would raise P0002 / ErrNoRows).
+            RAISE EXCEPTION 'insert or update on table "agent_connectors" violates foreign key constraint'
+                USING ERRCODE = '23503';
+        END IF;
     END IF;
     -- First row for (agent, connector) must be the default instance (column default is false).
     IF NOT EXISTS (
--- a/db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
+++ b/db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
@@ -52,6 +52,10 @@ CREATE TRIGGER trg_agent_connectors_before_insert
     FOR EACH ROW
     EXECUTE FUNCTION trg_agent_connectors_before_insert();
 
+-- agent_connector_credentials references this PK; drop FK before PK swap (re-added in next migration).
+ALTER TABLE agent_connector_credentials
+    DROP CONSTRAINT IF EXISTS agent_connector_credentials_agent_id_connector_id_fkey;
+
 ALTER TABLE agent_connectors DROP CONSTRAINT agent_connectors_pkey;
 
 ALTER TABLE agent_connectors
--- a/db/migrations/20260420145023_agent_connector_credentials_fk_to_instance.sql
+++ b/db/migrations/20260420145023_agent_connector_credentials_fk_to_instance.sql
@@ -14,8 +14,7 @@ WHERE acc.agent_id = ac.agent_id
 ALTER TABLE agent_connector_credentials
     ALTER COLUMN connector_instance_id SET NOT NULL;
 
-ALTER TABLE agent_connector_credentials
-    DROP CONSTRAINT agent_connector_credentials_agent_id_connector_id_fkey;
+-- Legacy FK already dropped in add_connector_instance_id_to_agent_connectors (required before PK swap).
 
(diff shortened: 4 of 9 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

## 🔧 CI / audit remediation

### Remediation round 1

- **ci.yml** — conclusion `failure` — Migration 20260420145022 failed: cannot DROP agent_connectors_pkey while agent_connector_credentials still had FK to (agent_id, connector_id). Dropped that FK at the start of the PK swap in migration 1; migration 2 no longer drops it (already gone). (`98f9aae9`)

### Remediation round 2

- **ci.yml** — conclusion `failure` — TestEnableAgentConnector_ConnectorNotFound failed: BEFORE INSERT trigger used STRICT lookup for connector name; missing connector raised P0002 and was mapped to agent_not_found. Trigger now raises ERRCODE 23503 so API returns connector_not_found. (`620971e9`)

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 6 (0 critical, 4 high, 1 medium, 1 low)

Phase 1 of the connector-instance work is thoughtfully structured: surrogate UUID + label + partial-unique default per `(agent, connector)`, backfill on existing rows, and BEFORE INSERT triggers to keep legacy `INSERT` shapes working. The main concerns are a couple of subtle correctness/forward-compat bugs in `db/agent_connectors.go` and a regression in the `AgentConnectorErrConnectorNotFound` error mapping caused by the new trigger. There's also a backfill edge case that could fail migration if `agent_connector_credentials.approver_id` ever drifts from its matching `agent_connectors.approver_id` row (the old FK didn't require this match). Nothing here is a merge-blocker on its own, but the `DisableAgentConnector` / `EnableAgentConnector` reads become latently incorrect as soon as Phase 2 introduces a second instance, so worth fixing now while the context is fresh.

### What Looks Good
- Clean separation of schema change + backfill + back-compat trigger in each migration; easy to read end-to-end.
- Partial unique index `WHERE is_default` is the right shape to enforce "exactly one default per pair".
- Down migrations are symmetric and restore the prior PK/FK/index topology — rare and appreciated.
- Backfill of `label` from `connectors.name` and marking all existing rows as default keeps legacy callers working without app-code churn in this PR.
- The nullable `standing_approvals.connector_instance_id` with "NULL = type-wide (legacy)" semantics is a clean way to stage Phase 4.

---

---

db/agent_connectors.go
```diff
@@ -140,7 +140,7 @@ func DisableAgentConnector(ctx context.Context, db DBTX, agentID int64, approver
 		WITH deleted AS (
 			DELETE FROM agent_connectors
 			WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
-			RETURNING agent_id, connector_id
+			RETURNING agent_id, connector_id, connector_instance_id
```
chiedoclaude, line 143: **[High · Correctness]** `DELETE ... WHERE agent_id=$1 AND approver_id=$2 AND connector_id=$3` deletes **all instances** for that tuple. In Phase 1 only the backfilled default exists, so behavior is unchanged — but as soon as Phase 2+ lands and a second instance is created for the same `(agent, connector)`, this silently wipes every instance. The `RETURNING ... connector_instance_id` you added is never consumed, which suggests the intent was per-instance semantics that didn't get wired up.

**Why:** A latent bug that will ship broken the moment multi-instance is enabled — and the return shape (`DisableAgentConnectorResult` still has no `ConnectorInstanceID` field) won't let the caller tell what got deleted.

**Fix:** Either (a) scope Phase 1 explicitly to the default, e.g. `AND is_default` in the DELETE, and drop the unused `connector_instance_id` from `RETURNING`; or (b) take a `connectorInstanceID` parameter and filter on it. Option (a) is the smaller diff for Phase 1 and matches the PR's stated scope.

---

db/agent_connectors.go
```diff
@@ -99,7 +99,7 @@ func EnableAgentConnector(ctx context.Context, db DBTX, agentID int64, approverI
 			INSERT INTO agent_connectors (agent_id, approver_id, connector_id)
 			SELECT $1, $2, $3
 			WHERE EXISTS (SELECT 1 FROM agent_check)
-			ON CONFLICT (agent_id, connector_id) DO NOTHING
+			ON CONFLICT (agent_id, connector_id, label) DO NOTHING
```
chiedoclaude, line 102: **[High · Forward Compat]** Related, below on line 107-109: the UNION-ALL fallback `SELECT ... FROM agent_connectors ac WHERE ac.agent_id=$1 AND ac.connector_id=$3` is unconstrained on instance. Today only one row matches (backfilled default), but in Phase 2+ with multiple instances per `(agent, connector)`, this branch returns multiple rows — `QueryRow.Scan` silently takes the first and callers get a non-deterministic `enabled_at`.

**Why:** `ON CONFLICT (agent_id, connector_id, label) DO NOTHING` correctly keeps this INSERT idempotent on the default instance (the trigger fills `label` from `connectors.name`), but the fallback read path doesn't share the same scoping — so this pair becomes inconsistent as soon as multi-instance lands.

**Fix:** Add `AND ac.is_default` to the UNION-ALL branch so it always resolves to exactly one row, matching the "first insert is the default" invariant the trigger establishes.

---

db/migrations/20260420145022_add_connector_instance_id_to_agent_connectors.sql
```diff
@@ -0,0 +1,78 @@
...
+    ON agent_connectors (agent_id, connector_id)
+    WHERE is_default;
+
+-- Back-compat: allow INSERT with only (agent_id, approver_id, connector_id); fill label and default flag.
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION trg_agent_connectors_before_insert()
+RETURNS trigger AS $$
+BEGIN
+    IF NEW.label IS NULL THEN
+        SELECT name INTO STRICT NEW.label FROM connectors WHERE id = NEW.connector_id;
```
chiedoclaude: **[High · Correctness]** `SELECT name INTO STRICT NEW.label FROM connectors WHERE id = NEW.connector_id` — with `STRICT`, if `connector_id` doesn't exist in `connectors`, this raises `NO_DATA_FOUND` (SQLSTATE `P0002`) inside the BEFORE INSERT trigger, **before** the FK constraint check runs.

**Why:** `EnableAgentConnector` in `db/agent_connectors.go` (line 119) maps `PgCodeForeignKeyViolation` → `AgentConnectorErrConnectorNotFound`. With this trigger in place, an invalid `connector_id` now produces `P0002` instead of `23503`, so that mapping silently stops firing and callers get a raw SQL error. Any handler checking `AgentConnectorErrConnectorNotFound` (e.g., to return a 404 instead of a 500) regresses.

**Fix:** Use a non-STRICT `SELECT ... INTO` and let the FK constraint raise the canonical violation:

```sql
IF NEW.label IS NULL THEN
    SELECT name INTO NEW.label FROM connectors WHERE id = NEW.connector_id;
    -- If NEW.label is still NULL, the FK check below will reject with 23503.
END IF;
```

Or explicitly `RAISE foreign_key_violation` so the existing error mapping keeps working. A test that calls `EnableAgentConnector` with a bogus `connector_id` and asserts `AgentConnectorErrConnectorNotFound` would catch this.

---

db/migrations/20260420145023_agent_connector_credentials_fk_to_instance.sql
```diff
@@ -0,0 +1,74 @@
...
+
+UPDATE agent_connector_credentials acc
+SET connector_instance_id = ac.connector_instance_id
+FROM agent_connectors ac
+WHERE acc.agent_id = ac.agent_id
+  AND acc.connector_id = ac.connector_id
+  AND acc.approver_id = ac.approver_id;
+
+ALTER TABLE agent_connector_credentials
+    ALTER COLUMN connector_instance_id SET NOT NULL;
```
chiedoclaude, line 15: **[High · Migration Safety]** The backfill on lines 7–12 matches `acc.approver_id = ac.approver_id`, but the pre-PR FK on `agent_connector_credentials(agent_id, connector_id)` only enforced `(agent_id, connector_id)` — `approver_id` was not part of the old FK. If any existing `agent_connector_credentials` row has drifted from its matching `agent_connectors.approver_id` (for example, an approver was rotated, or there's any historical inconsistency), `connector_instance_id` stays NULL for that row and this `ALTER COLUMN ... SET NOT NULL` fails mid-migration.

**Why:** Migration failure on production data is hard to recover from (rollback requires Down, which needs its own clean state), and the symptom — a `null value in column ... violates not-null constraint` — doesn't point at the drifted row. Even if drift is unlikely today, the migration shouldn't depend on an invariant the old schema didn't enforce.

**Fix:** Either (a) add a pre-check that surfaces drift explicitly:

```sql
DO $$ DECLARE bad_count int;
BEGIN
  SELECT count(*) INTO bad_count FROM agent_connector_credentials
  WHERE connector_instance_id IS NULL;
  IF bad_count > 0 THEN
    RAISE EXCEPTION 'Cannot backfill %: rows in agent_connector_credentials have no matching agent_connectors row (check approver_id drift)', bad_count;
  END IF;
END $$;
```

…run before the `SET NOT NULL`, so the migration fails with a clear diagnostic instead of a confusing NOT NULL violation; or (b) fall back to a two-step backfill (first match on `(agent_id, approver_id, connector_id)`, then match any remaining rows on `(agent_id, connector_id)` only, updating `approver_id` to align).

---

db/migrations/20260420145024_standing_approvals_add_connector_instance_id.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- NULL = type-wide standing approval (legacy); non-null = instance-scoped (Phase 4).
+
+ALTER TABLE standing_approvals
+    ADD COLUMN connector_instance_id uuid;
+
+CREATE INDEX idx_standing_approvals_agent_action_status_connector_instance
+    ON standing_approvals (agent_id, action_type, status, connector_instance_id);
```
chiedoclaude, line 8: **[Medium · Maintainability]** This index is a superset of the existing `idx_standing_approvals_agent_action_status` on `(agent_id, action_type, status)` — Postgres can satisfy any query the old index serves using this one (the old index is only selective for "queries that don't touch instance"), so you're paying double write cost on `standing_approvals` inserts/updates without a clear benefit until Phase 4 queries land.

**Why:** Redundant indexes bloat the table and slow writes; they also confuse future readers (which one is authoritative?).

**Fix:** Either drop `idx_standing_approvals_agent_action_status` in the same migration (the new 4-col index serves the same prefix lookups), or add a comment explaining why both are intentionally kept through Phase 4. If the old one is deleted, update the corresponding `RequireIndex` in `db/standing_approvals_test.go`.

---

db/agent_connector_credentials_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+func TestAgentConnectorCredentialsSchema(t *testing.T) {
+	t.Parallel()
+	tx := testhelper.SetupTestDB(t)
+
+	testhelper.RequireColumns(t, tx, "agent_connector_credentials", []string{
+		"id", "agent_id", "connector_id", "approver_id", "connector_instance_id",
+		"credential_id", "oauth_connection_id", "created_at",
+	})
+}
```
chiedoclaude, line 17: **[Medium · Testing]** Schema tests added here and in `agent_connectors_test.go` only check column presence — they won't catch regressions in the new semantic behavior this PR introduces. Nothing exercises:

- `trg_agent_connectors_before_insert` filling `label` from `connectors.name` on an INSERT that omits it
- `trg_agent_connectors_before_insert` setting `is_default=true` for the first row of a `(agent, connector)` pair and `false` for subsequent rows
- `trg_agent_connector_credentials_before_insert` resolving omitted `connector_instance_id` to the default instance's UUID
- `uq_agent_connectors_default_per_pair` rejecting a second `is_default=true` row for the same pair
- The FK `agent_connector_credentials_agent_instance_fkey` cascading on `agent_connectors` delete
- `EnableAgentConnector` still returning `AgentConnectorErrConnectorNotFound` for an invalid `connector_id` (regression risk from the `INTO STRICT` issue noted elsewhere)

**Why:** Each of these is a behavior the rest of Phase 1–4 will rely on; without tests they can silently regress in later migrations (e.g., someone "cleans up" a trigger, or a future column default shadows `is_default`).

**Fix:** Add integration tests in `db/` that INSERT legacy-shape rows and assert the triggered behavior. The existing `testhelper` already has `InsertAgentConnector` / `InsertConnector` — should be ~30 lines per case.

---

chiedoclaude (commented):
## Review Round 1/6 — Inline Findings

Detailed inline comments below complementing the earlier top-level summary.

**Inline findings:** 5 (0 critical, 3 high, 2 medium, 0 low)

### High
- `DisableAgentConnector` deletes all instances; `RETURNING connector_instance_id` unused — latent bug for Phase 2+
- `EnableAgentConnector` fallback `SELECT` unconstrained on instance — non-deterministic in Phase 2+
- Trigger's `SELECT ... INTO STRICT` regresses `AgentConnectorErrConnectorNotFound` error mapping
- Backfill in `20260420145023` assumes `(agent_id, approver_id, connector_id)` match that the old FK didn't enforce — risks mid-migration `SET NOT NULL` failure

### Medium
- New 4-col index on `standing_approvals` is a superset of existing 3-col index — redundant writes
- Schema tests only check column presence; triggers / partial-unique / FK cascade aren't exercised

---

---

chiedoclaude:
## Code Review Complete

**PR:** #963 — feat(db): connector instance schema + backfill (issue #962 phase 1)
**Author:** @chiedocursor
**Rounds completed:** 3/6 (stopped early)
**Reason for stopping:** no activity for 10 minutes after last push

### Outstanding Issues

From Round 1, five findings remain open:

- **[High · Correctness]** `DisableAgentConnector` in `db/agent_connectors.go` — `DELETE` matches all instances for `(agent_id, approver_id, connector_id)`; `RETURNING connector_instance_id` is unused. Works today (single backfilled default) but silently wrong as soon as Phase 2+ adds a second instance. Either add `AND is_default` for Phase 1 or take a `connectorInstanceID` arg.
- **[High · Forward Compat]** `EnableAgentConnector` UNION-ALL fallback (line 107–109) — unconstrained on instance; `QueryRow.Scan` returns a non-deterministic row in Phase 2+. Fix: add `AND ac.is_default` to match the trigger's "first insert is the default" invariant.
- **[High · Migration Safety]** Backfill in `20260420145023` matches `acc.approver_id = ac.approver_id`, an invariant the old FK didn't enforce. Any drift fails `SET NOT NULL` mid-migration with a confusing error. Add a diagnostic pre-check or two-step fallback.
- **[Medium · Maintainability]** New 4-col `standing_approvals` index supersets the existing 3-col one. Either drop the old index or document why both are intentionally retained through Phase 4.
- **[Medium · Testing]** Schema tests only check column presence. Nothing exercises: the back-compat triggers (label fill, default-instance fill), the partial unique `WHERE is_default` index, FK cascade on instance delete, or `EnableAgentConnector` → `AgentConnectorErrConnectorNotFound` for bogus `connector_id`. The last one is now easy to assert since the trigger was just fixed to emit SQLSTATE 23503 — a direct regression test there would lock in the fix.

### Review History

| 1 | 5 | Initial review: 3 high, 2 medium |
| 2 | 0 | CI-fix commit `98f9aae` (FK drop reorder); no Round 1 findings addressed |
| 3 | 0 | `620971e` fixed the `INTO STRICT` → `P0002` issue with an explicit `RAISE ERRCODE = '23503'`; other Round 1 findings remain open |

### Overall Assessment

Well-structured Phase 1 migration with good choices (surrogate UUID, partial unique default, backfill + back-compat triggers, symmetric down migrations). The trigger fix in round 3 is a clean response to the error-code regression. **Not merge-blocking** as-is because Phase 1 only has one default instance per `(agent, connector)`, so the `DisableAgentConnector` / `EnableAgentConnector` scoping issues don't manifest today — but they're loaded weapons waiting for Phase 2 to introduce a second instance, so cheaper to fix now (2–3 line changes) than to remember later. The backfill approver_id concern is the one worth a defensive guard even if drift is believed to be absent.

---

---

PR #964, opened 2026-04-20
Phase 2: Backend connector instance CRUD and API routes

chiedocursor:
## Summary

Implements **Phase 2** of #962: database helpers for connector instances, new REST routes under `/agents/{agent_id}/connectors/{connector_id}/instances`, credential and metadata handlers scoped by `instance_id`, legacy routes targeting the **default** instance, standing approval struct/insert support for `connector_instance_id`, and `FindActiveStandingApprovalsForAgent` filtering when a non-empty instance ID is passed (Phase 4 can thread the resolved instance through).

OpenAPI: new paths and schemas, `ConnectorInstanceId` path parameter, legacy `/credential` operations marked **deprecated** in descriptions, new error codes `connector_instance_required` and `connector_instance_not_found`, bundled spec updated. `google/uuid` is now a direct module dependency for instance selector parsing.

### Developer

- [x] `make bundle` + frontend `tsc -b --noEmit` after spec change
- [x] `make test-frontend` — passed (1008 tests)
- [ ] `go test ./db/...` and `go test ./api/...` — **not run in this environment** (no PostgreSQL on localhost; `go test` also requires toolchain/modules consistent with CI). New tests: `db/agent_connector_instances_test.go`, `api/agent_connector_instances_test.go`.

### Manual Testing

- [ ] Create a second Slack instance via `POST .../instances`; list shows two rows with distinct UUIDs
- [ ] Legacy `GET/PUT/DELETE .../credential` still affects only the default instance when multiple instances exist
- [ ] `PATCH .../instances/{id}` duplicate label returns 409
- [ ] `DELETE` non-default instance removes credential row (cascade) and revokes instance-scoped standing approvals

Refs #962

<div><a href="https://cursor.com/agents/bc-0cbeaa98-b16c-41dc-8889-48ee9c452248"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0cbeaa98-b16c-41dc-8889-48ee9c452248"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/api/agent_connector_instances.go
+++ b/api/agent_connector_instances.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/supersuit-tech/permission-slip/connectors"
 	"github.com/supersuit-tech/permission-slip/db"
 )
@@ -127,6 +128,10 @@ func handleCreateAgentConnectorInstance(deps *Deps) http.HandlerFunc {
 			Label:       req.Label,
 		})
 		if err != nil {
+			if errors.Is(err, db.ErrAgentConnectorInstanceLabelRequired) || errors.Is(err, db.ErrAgentConnectorInstanceLabelTooLong) {
+				RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "label must be 1–256 characters"))
+				return
+			}
 			var acErr *db.AgentConnectorError
 			if errors.As(err, &acErr) {
 				switch acErr.Code {
@@ -135,6 +140,9 @@ func handleCreateAgentConnectorInstance(deps *Deps) http.HandlerFunc {
 					return
 				case db.AgentConnectorErrConnectorNotFound:
 					RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorNotFound, "Connector not found"))
+					return
+				case db.AgentConnectorErrConnectorNotEnabled:
+					RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "Enable this connector for the agent before adding another instance"))
 					return
 				}
 			}
@@ -159,7 +167,27 @@ func parsePathInstanceID(w http.ResponseWriter, r *http.Request) (string, bool) 
 		RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "instance_id is required"))
 		return "", false
 	}
+	if _, err := uuid.Parse(id); err != nil {
+		RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "instance_id must be a valid UUID"))
+		return "", false
+	}
 	return id, true
+}
+
+// requireAgentConnectorInstance loads the instance or writes 404/500 and returns nil, false on failure.
+func requireAgentConnectorInstance(w http.ResponseWriter, r *http.Request, deps *Deps, agentID int64, userID, connectorID, instanceID string) (*db.AgentConnectorInstance, bool) {
+	inst, err := db.GetAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID)
+	if err != nil {
+		log.Printf("[%s] GetAgentConnectorInstance: %v", TraceID(r.Context()), err)
+		CaptureError(r.Context(), err)
+		RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to verify connector instance"))
+		return nil, false
+	}
+	if inst == nil {
+		RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+		return nil, false
+	}
+	return inst, true
 }
 
 func handleGetAgentConnectorInstance(deps *Deps) http.HandlerFunc {
@@ -182,15 +210,8 @@ func handleGetAgentConnectorInstance(deps *Deps) http.HandlerFunc {
 			return
 		}
 
-		inst, err := db.GetAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID)
-		if err != nil {
-			log.Printf("[%s] GetAgentConnectorInstance: %v", TraceID(r.Context()), err)
-			CaptureError(r.Context(), err)
-			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to get connector instance"))
-			return
-		}
-		if inst == nil {
-			RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+		inst, instOK := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID)
+		if !instOK {
 			return
 		}
 		RespondJSON(w, http.StatusOK, toAgentConnectorInstanceResponse(*inst))
@@ -224,6 +245,10 @@ func handlePatchAgentConnectorInstance(deps *Deps) http.HandlerFunc {
 
 		inst, err := db.RenameAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID, req.Label)
 		if err != nil {
+			if errors.Is(err, db.ErrAgentConnectorInstanceLabelRequired) || errors.Is(err, db.ErrAgentConnectorInstanceLabelTooLong) {
+				RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "label must be 1–256 characters"))
+				return
+			}
 			var instErr *db.AgentConnectorInstanceError
 			if errors.As(err, &instErr) && instErr.Code == db.AgentConnectorInstanceErrDuplicateLabel {
 				RespondError(w, r, http.StatusConflict, Conflict(ErrConstraintViolation, "An instance with this label already exists"))
@@ -304,15 +329,7 @@ func handleAssignAgentConnectorInstanceCredential(deps *Deps) http.HandlerFunc {
 			return
 		}
 
-		inst, err := db.GetAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID)
-		if err != nil {
-			log.Printf("[%s] AssignInstanceCredential: get instance: %v", TraceID(r.Context()), err)
-			CaptureError(r.Context(), err)
-			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to verify connector instance"))
-			return
-		}
-		if inst == nil {
-			RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+		if _, ok := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID); !ok {
 			return
 		}
 
@@ -377,15 +394,7 @@ func handleRemoveAgentConnectorInstanceCredential(deps *Deps) http.HandlerFunc {
 			return
 		}
 
-		inst, err := db.GetAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID)
-		if err != nil {
-			log.Printf("[%s] RemoveInstanceCredential: get instance: %v", TraceID(r.Context()), err)
-			CaptureError(r.Context(), err)
-			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to verify connector instance"))
-			return
-		}
-		if inst == nil {
-			RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+		if _, ok := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID); !ok {
 			return
 		}
 
@@ -424,15 +433,7 @@ func handleGetAgentConnectorInstanceCredential(deps *Deps) http.HandlerFunc {
 			return
 		}
 
-		inst, err := db.GetAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID)
-		if err != nil {
-			log.Printf("[%s] GetInstanceCredential: get instance: %v", TraceID(r.Context()), err)
-			CaptureError(r.Context(), err)
-			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to verify connector instance"))
-			return
-		}
-		if inst == nil {
-			RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+		if _, ok := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID); !ok {
 			return
 		}
 
@@ -478,6 +479,10 @@ func handleListAgentConnectorInstanceChannels(deps *Deps) http.HandlerFunc {
 			return
 		}
 		if !requireAgentOwnership(w, r, deps, agentID, userID) {
+			return
+		}
+
+		if _, ok := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID); !ok {
 			return
 		}
 
@@ -569,6 +574,10 @@ func handleListAgentConnectorInstanceCalendars(deps *Deps) http.HandlerFunc {
 			return
 		}
 
+		if _, ok := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID); !ok {
+			return
+		}
+
 		if deps.Connectors == nil {
 			RespondError(w, r, http.StatusServiceUnavailable, ServiceUnavailable("Connectors are not available"))
 			return
@@ -654,6 +663,10 @@ func handleListAgentConnectorInstanceUsers(deps *Deps) http.HandlerFunc {
 			return
 		}
 		if !requireAgentOwnership(w, r, deps, agentID, userID) {
+			return
+		}
+
+		if _, ok := requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID); !ok {
 			return
 		}
 
--- a/api/agent_connector_instances.go
+++ b/api/agent_connector_instances.go
@@ -1,0 +1,725 @@
+package api
+
+import (
+	"errors"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+	"github.com/supersuit-tech/permission-slip/db"
+)
+
+type agentConnectorInstanceResponse struct {
+	ConnectorInstanceID string    `json:"connector_instance_id"`
+	AgentID             int64     `json:"agent_id"`
+	ConnectorID         string    `json:"connector_id"`
+	Label               string    `json:"label"`
+	IsDefault           bool      `json:"is_default"`
+	EnabledAt           time.Time `json:"enabled_at"`
+}
+
+type agentConnectorInstanceListResponse struct {
+	Data []agentConnectorInstanceResponse `json:"data"`
+}
+
+type createAgentConnectorInstanceRequest struct {
+	Label string `json:"label"`
+}
+
+type patchAgentConnectorInstanceRequest struct {
+	Label string `json:"label"`
+}
+
+func init() {
+	RegisterRouteGroup(RegisterAgentConnectorInstanceRoutes)
+}
+
+// RegisterAgentConnectorInstanceRoutes registers multi-instance connector endpoints.
+func RegisterAgentConnectorInstanceRoutes(mux *http.ServeMux, deps *Deps) {
+	requireProfile := RequireProfile(deps)
+	base := "/agents/{agent_id}/connectors/{connector_id}/instances"
+	mux.Handle("GET "+base, requireProfile(handleListAgentConnectorInstances(deps)))
+	mux.Handle("POST "+base, requireProfile(handleCreateAgentConnectorInstance(deps)))
+
+	inst := base + "/{instance_id}"
+	mux.Handle("GET "+inst, requireProfile(handleGetAgentConnectorInstance(deps)))
+	mux.Handle("PATCH "+inst, requireProfile(handlePatchAgentConnectorInstance(deps)))
+	mux.Handle("DELETE "+inst, requireProfile(handleDeleteAgentConnectorInstance(deps)))
+
+	mux.Handle("PUT "+inst+"/credential", requireProfile(handleAssignAgentConnectorInstanceCredential(deps)))
+	mux.Handle("DELETE "+inst+"/credential", requireProfile(handleRemoveAgentConnectorInstanceCredential(deps)))
+	mux.Handle("GET "+inst+"/credential", requireProfile(handleGetAgentConnectorInstanceCredential(deps)))
+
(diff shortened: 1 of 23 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 7 (0 critical, 3 high, 3 medium, 1 low)

Phase 2 wires up a clean multi-instance model with deprecations on the legacy routes and a sensible backfill path. The biggest concerns are around durability and input validation:

- `DeleteAgentConnectorInstance` runs SELECT → UPDATE standing_approvals → DELETE agent_connectors with no transaction, so a partial failure leaves standing approvals spuriously revoked.
- A malformed UUID in any `/instances/{instance_id}` route surfaces as 500 (PostgreSQL `22P02`) instead of 400/404, and noisily lands in Sentry.
- `handleDisableAgentConnector` only collects the *default* instance's credential binding when `delete_credentials=true`, so vault secrets for non-default instances are orphaned on full-connector disable.

Medium items: the channels/calendars/users instance handlers skip the `GetAgentConnectorInstance` existence check the credential handlers do (inconsistent + confusing error); `CreateAgentConnectorInstance` can produce a "default" instance when the connector was never enabled; `label` has no length bound. Plus a few testing gaps worth closing before merge.

### What Looks Good
- Clean split between legacy-default-scoped helpers (`GetAgentConnectorCredential`, `UpsertAgentConnectorCredential`, `DeleteAgentConnectorCredential`) and new `*ByInstance` variants — minimal disruption to callers.
- `FindActiveStandingApprovalsForAgent` instance filter correctly keeps NULL (type-wide) approvals matching, preserving back-compat; the DB test covers the three cases.
- Thoughtful OpenAPI: `deprecated: true` on legacy credential routes plus new error codes makes the migration story explicit for clients.
- Proper cascade-revoke of instance-scoped standing approvals on instance delete, with a dedicated test.

---

---

chiedoclaude (commented):
Inline findings for Round 1 (see the main review summary for context).

---

db/agent_connector_instances.go
```diff
@@ -0,0 +1,290 @@
...
+		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3 AND connector_instance_id = $4::uuid`,
+		agentID, approverID, connectorID, connectorInstanceID,
+	)
+	if err != nil {
+		return err
+	}
+	if tag.RowsAffected() == 0 {
+		return &AgentConnectorInstanceError{Code: AgentConnectorInstanceErrNotFound}
+	}
+	return nil
```
chiedoclaude, line 304: **[High · Correctness]** This function runs three statements (SELECT `is_default`, UPDATE `standing_approvals`, DELETE `agent_connectors`) without a transaction, and the caller (`handleDeleteAgentConnectorInstance`) passes `deps.DB`, not a tx. If the DELETE fails after the UPDATE commits — connection drop, FK violation, Postgres restart — standing approvals are left `revoked` while the instance still exists, and a retry revokes any active approvals attached to the still-live instance a second time.

**Why:** Silent data corruption. The user sees a 500 but their approvals have already flipped to revoked and their instance is still there; retrying the delete doesn't get it back.

**Fix:** Wrap in a transaction, consistent with `handleDisableAgentConnector`:

```go
func DeleteAgentConnectorInstance(ctx context.Context, dbtx DBTX, agentID int64, approverID, connectorID, connectorInstanceID string) error {
    tx, owned, err := BeginOrContinue(ctx, dbtx)
    if err != nil {
        return err
    }
    if owned {
        defer RollbackTx(ctx, tx) //nolint:errcheck
    }
    // ... existing SELECT / UPDATE / DELETE against tx ...
    if owned {
        return CommitTx(ctx, tx)
    }
    return nil
}
```

Same pattern would also let you swap the SELECT + DELETE for a single `DELETE ... WHERE NOT is_default RETURNING is_default` to close the TOCTOU between the is_default read and the delete.

---

api/agent_connector_instances.go
```diff
@@ -0,0 +1,725 @@
...
+}
+
+func parsePathInstanceID(w http.ResponseWriter, r *http.Request) (string, bool) {
+	id := r.PathValue("instance_id")
+	if id == "" {
+		RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "instance_id is required"))
+		return "", false
+	}
+	return id, true
+}
```
chiedoclaude, line 175: **[High · Code Quality]** `parsePathInstanceID` only rejects empty string — any non-empty non-UUID value (e.g. `/instances/foo`) reaches the DB query that casts `$N::uuid`, Postgres raises `22P02 invalid_text_representation`, and the handler returns 500 and fires `CaptureError` to Sentry.

**Why:** User-controlled malformed input should not be a 500 or produce Sentry noise. Every endpoint under `/instances/{instance_id}` (GET, PATCH, DELETE, plus the credential/channels/calendars/users proxies) is affected because they all go through this helper.

**Fix:** Validate with `uuid.Parse` here and return 400 (or 404 if you want opaque "not found"):

```go
func parsePathInstanceID(w http.ResponseWriter, r *http.Request) (string, bool) {
    id := r.PathValue("instance_id")
    if id == "" {
        RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "instance_id is required"))
        return "", false
    }
    if _, err := uuid.Parse(id); err != nil {
        RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
        return "", false
    }
    return id, true
}
```

(404 matches how you'd want this to behave in practice — an unparseable UUID is indistinguishable from "doesn't exist".)

---

api/agent_connectors.go
```diff
@@ -159,13 +159,22 @@ func handleDisableAgentConnector(deps *Deps) http.HandlerFunc {
...
+			if defaultInst != nil {
+				credBinding, err = db.GetAgentConnectorCredentialByInstance(r.Context(), tx, agentID, connectorID, defaultInst.ConnectorInstanceID)
+				if err != nil {
+					log.Printf("[%s] DisableAgentConnector: get credential: %v", TraceID(r.Context()), err)
+					CaptureError(r.Context(), err)
+					RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to disable connector"))
+					return
+				}
+			}
 		}
```
chiedoclaude, line 169: **[High · Correctness]** When `delete_credentials=true`, this block only fetches the *default* instance's binding. Since `DisableAgentConnector` cascade-deletes every instance (and their `agent_connector_credentials` rows) and `handleDisableAgentConnector` only calls `deleteCredentialAndVault` for the single `credBinding` collected here, the `credentials` rows and vault secrets bound to non-default instances are silently orphaned.

**Why:** Vault secrets and credential rows leak on every multi-instance connector disable. The user explicitly asked for cleanup (`delete_credentials=true`) and doesn't get it.

**Fix:** Collect all bindings across all instances for the agent+connector, dedupe `CredentialID`s (the same credential may be bound to multiple instances), and call `deleteCredentialAndVault` for each. Something like:

```go
bindings, err := db.ListAgentConnectorCredentialsByAgentConnector(r.Context(), tx, agentID, userID, connectorID)
// ... then iterate unique credBinding.CredentialID values below.
```

If adding `ListAgentConnectorCredentialsByAgentConnector` is out of scope for Phase 2, at minimum file a follow-up issue and document the leak in the OpenAPI `delete_credentials` description — today the docs imply cleanup is complete.

---

api/agent_connector_instances.go
```diff
@@ -0,0 +1,725 @@
...
+		}
+		reqCreds, err := db.GetRequiredCredentialsByActionType(r.Context(), deps.DB, listAction)
+		if err != nil {
+			log.Printf("[%s] ListAgentConnectorInstanceChannels required creds: %v", TraceID(r.Context()), err)
+			CaptureError(r.Context(), err)
+			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to look up connector credentials"))
+			return
+		}
+
+		creds, err := resolveCredentialsForConnectorInstance(r.Context(), deps, agentID, userID, listAction, connectorID, instanceID, reqCreds)
```
chiedoclaude, line 520: **[Medium · UX / Consistency]** `handleListAgentConnectorInstanceChannels` (and its Calendars/Users twins below) skip the `GetAgentConnectorInstance` existence check that every credential handler in this same file performs. If the caller passes a valid UUID for an instance that belongs to another agent (or doesn't exist), the flow falls through to `GetAgentConnectorCredentialByInstance`, returns nil, and surfaces a `ValidationError` "no credential assigned — go to the agent's connector settings and assign a credential."

**Why:** Misleading error message (credential vs. instance) and inconsistent 404 behaviour across sibling endpoints. Not a security leak — `agent_id` scoping prevents cross-tenant data access — but the UX is noticeably worse than the credential routes above.

**Fix:** Add the same `GetAgentConnectorInstance` + nil-check upfront that `handleAssignAgentConnectorInstanceCredential` uses (lines 307-317). Consider extracting it into a small helper `requireAgentConnectorInstance(w, r, deps, agentID, userID, connectorID, instanceID) (*db.AgentConnectorInstance, bool)` since it's repeated six times in this file.

---

db/agent_connector_instances.go
```diff
@@ -0,0 +1,290 @@
...
+	inst, err := scanAgentConnectorInstance(row)
+	if err != nil {
+		var pgErr *pgconn.PgError
+		if errors.As(err, &pgErr) && pgErr.Code == PgCodeUniqueViolation {
+			return nil, &AgentConnectorInstanceError{Code: AgentConnectorInstanceErrDuplicateLabel}
+		}
+		return nil, err
+	}
+	return inst, nil
+}
```
chiedoclaude, line 122: **[Medium · API contract]** This function inserts `is_default = false`, but the DB trigger `trg_agent_connectors_before_insert` flips it to `true` when no default exists for the (agent, connector) pair. So calling `POST /instances` on an agent where the connector was *never* enabled silently creates the caller's label as the default instance — bypassing the enable flow and hiding the fact from the client (the returned row shows `is_default: true`, which is not what the doc comment on line 51 describes).

**Why:** API contract drift: the OpenAPI description for this endpoint says "adds further instances" implying a default already exists. A client building on that assumption will write surprising UX (e.g., letting a user name a "secondary" instance that silently becomes their only/default one).

**Fix:** Add a default-existence precondition before the INSERT:

```go
defaultInst, err := GetDefaultAgentConnectorInstance(ctx, db, p.AgentID, p.ApproverID, p.ConnectorID)
if err != nil {
    return nil, err
}
if defaultInst == nil {
    return nil, &AgentConnectorError{Code: AgentConnectorErrConnectorNotEnabled}
}
```

(Map to 409 or 400 in the handler.)

---

spec/openapi/components/schemas/agent_connector_instances.yaml
```diff
@@ -0,0 +1,57 @@
...
+        $ref: '#/AgentConnectorInstance'
+
+CreateAgentConnectorInstanceRequest:
+  type: object
+  required:
+    - label
+  properties:
+    label:
+      type: string
+      description: Display label for the new instance (unique per agent and connector).
```
chiedoclaude, line 50: **[Medium · Input validation]** `label` has no `maxLength` constraint here (or in the `PatchAgentConnectorInstanceRequest` below), and the DB column is plain `text`. `CreateAgentConnectorInstance` only trims whitespace — a client can POST a multi-MB label and it will be stored and returned on every List call.

**Why:** Unbounded user-controlled text stored in a hot-read path (listed on every agent settings page load) is an easy DoS / storage inflation vector, and the OpenAPI-generated TypeScript client won't warn about it either.

**Fix:** Pick a reasonable bound and apply it in both places:

```yaml
label:
  type: string
  minLength: 1
  maxLength: 128
  description: ...
```

Enforce the same limit in `CreateAgentConnectorInstance`/`RenameAgentConnectorInstance` (`if len(label) > 128 { return ..., errors.New("label too long") }`) so non-spec-validated callers are still rejected with a 400.

---

api/agent_connector_instances_test.go
```diff
@@ -0,0 +1,129 @@
...
+		t.Fatal(err)
+	}
+	r2 := authenticatedRequestWithBody(t, http.MethodPatch,
+		fmt.Sprintf("/agents/%d/connectors/%s/instances/%s", agentID, connID, created.ConnectorInstanceID), uid, patchPayload)
+	w2 := httptest.NewRecorder()
+	router.ServeHTTP(w2, r2)
+	if w2.Code != http.StatusConflict {
+		t.Fatalf("expected 409, got %d: %s", w2.Code, w2.Body.String())
+	}
+}
```
chiedoclaude, line 129: **[Low · Test coverage]** API tests only cover List-default, Create-second, and Patch-duplicate-label. For a handler file this large, please add coverage for:

- `DELETE /instances/{id}` — happy path for non-default and 400 for default.
- `GET /instances/{id}` returning 404 for (a) unknown UUID and (b) an instance that belongs to another user's agent.
- `PUT/GET/DELETE .../credential` — at least the assign + get round-trip.
- Invalid/malformed `instance_id` (ties into the 500-vs-400 comment in `agent_connector_instances.go`).

The db-level tests cover the cascade-revoke case nicely; the gap is in the HTTP surface.

---

chiedoclaude (commented):
## Review Round 1/6 — Inline Findings

**Findings:** 7 (0 critical, 3 high, 3 medium, 1 low)

Inline comments cover the main themes from the Round 1 summary: a non-transactional multi-statement delete, a vault-secret cleanup gap on legacy `disable` with non-default instances, a 500 on malformed UUIDs, an implicit-default creation path, missing label length validation, a few handlers that skip instance existence checks, and test coverage gaps for the new endpoints. See each inline comment for the consequence and a concrete fix.

### What Looks Good
- Clean separation between legacy type-wide and instance-scoped routes, with legacy paths correctly resolving the default instance.
- Thoughtful backfill strategy via the pre-insert trigger (from Phase 1) plus explicit `is_default` semantics on the new helpers.
- Good test coverage of the happy path for list/create/rename plus the standing-approval revoke-on-delete flow.
- Deprecation markers on legacy credential routes make the migration path discoverable in the spec.

---

---

chiedocursor:
⚠️ **Auto-merge failed.** The `--automerge` flag was set, but the merge could not be completed. Please merge manually.

db/agent_connector_instances.go
chiedocursor: Fixed: DeleteAgentConnectorInstance now uses BeginOrContinue and runs SELECT → UPDATE standing_approvals → DELETE agent_connectors in one transaction (commit f104b4dc).

api/agent_connector_instances.go
chiedocursor: Fixed: parsePathInstanceID validates with uuid.Parse and returns 400 invalid_request for non-UUID instance_id (commit f104b4dc).

api/agent_connectors.go
chiedocursor: Fixed: delete_credentials=true now lists all bindings via ListAgentConnectorCredentialsForAgentConnector and deletes each static credential vault secret before DisableAgentConnector cascade (commit f104b4dc).

api/agent_connector_instances.go
chiedocursor: Fixed: instance channels/calendars/users handlers now call requireAgentConnectorInstance before resolving credentials (commit f104b4dc).

db/agent_connector_instances.go
chiedocursor: Fixed: CreateAgentConnectorInstance requires an existing agent_connectors row (AgentConnectorErrConnectorNotEnabled) so we do not implicitly create a default via trigger; DB test added (commit f104b4dc).

spec/openapi/components/schemas/agent_connector_instances.yaml
chiedocursor: Fixed: OpenAPI label fields have minLength 1 and maxLength 256 (commit f104b4dc).

api/agent_connector_instances_test.go
chiedocursor: Fixed: added API tests for invalid UUID (400), DELETE second instance (204), plus DB test for create when connector not enabled (commit f104b4dc).

api/agent_connector_instances_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	rList := authenticatedRequest(t, http.MethodGet, fmt.Sprintf("/agents/%d/connectors/%s/instances", agentID, connID), uid)
+	wList := httptest.NewRecorder()
+	router.ServeHTTP(wList, rList)
+	list := decodeInstanceList(t, wList.Body.Bytes())
+	if len(list.Data) != 1 {
+		t.Fatalf("after delete expected 1 instance, got %d", len(list.Data))
+	}
+}
```
chiedoclaude, line 192: **[Low · Test Coverage]** Three residual gaps in this test file after this round:

1. The updated disable-with-`delete_credentials=true` behavior (now iterates `ListAgentConnectorCredentialsForAgentConnector` and calls `deleteCredentialAndVault` for every binding) has no regression test. That was the Round 1 High-severity fix — without a test, a future refactor could silently revert to single-instance cleanup and leak non-default vault secrets again.
2. No API tests for the instance credential routes (`PUT/GET/DELETE .../instances/{id}/credential`). The shared `validateAssignCredentialRequest` helper is covered for the legacy path but not for the new instance path.
3. No cross-tenant access tests (e.g., user B requests `GET /agents/{userA_agent}/connectors/{id}/instances/{id}` → expect 404). `requireAgentOwnership` is exercised by legacy routes, but re-verifying it on the new routes would guard against accidental middleware drift.

**Why:** A single test of the form "create two instances each with a distinct credential → `POST /disable?delete_credentials=true` → assert both `vault.DeleteSecret` calls fired" would lock in the fix for #964's most impactful change. The other two are lower priority but round out the coverage for this phase.

**Fix:** Add three tests:
```go
TestDisableAgentConnector_DeleteCredentialsRemovesAllInstances  // covers the Round 1 high-sev fix
TestAssignAgentConnectorInstanceCredential_InstanceRoute       // basic happy path on the instance credential route
TestGetAgentConnectorInstance_CrossTenant404                   // user B hits user A's agent → 404
```
These can use the existing `testhelper.InsertUserWithAgent` and a fake `deps.Vault` recorder to assert deletion calls.

---

chiedoclaude (commented):
## Review Round 2/6

**Findings:** 1 (0 critical, 0 high, 0 medium, 1 low)

### Changes Since Last Round
New commit `f104b4d` — "fix: address PR review — instance delete tx, UUID validation, multi-instance disable". Addressed 6 of the 7 Round 1 findings:

- **[High · Correctness]** `DeleteAgentConnectorInstance` now wraps the SELECT + standing-approval revoke + DELETE in `BeginOrContinue`/`CommitTx`. ✓
- **[High · Code Quality]** `parsePathInstanceID` now validates with `uuid.Parse`; malformed UUID returns 400. ✓
- **[High · Correctness]** `handleDisableAgentConnector` now uses new `ListAgentConnectorCredentialsForAgentConnector` and loops over every binding. Non-default vault secrets are no longer leaked. ✓
- **[Medium · UX]** New `requireAgentConnectorInstance` helper DRYs up the existence check and is now applied to channels/calendars/users/credential handlers. ✓
- **[Medium · API contract]** `CreateAgentConnectorInstance` now returns `connector_not_enabled` (400) if no row exists for the agent+connector. ✓
- **[Medium · Validation]** Schema now has `minLength: 1` / `maxLength: 256` on label; Go enforces with `utf8.RuneCountInString` against `MaxAgentConnectorInstanceLabelLen = 256`. Consistent with the `text` DB column. ✓
- **[Low · Tests]** Added: invalid-UUID 400, delete-second-instance 204, create-without-enabled-connector. Partial.

### Remaining Issues
One low-severity test-coverage gap — see inline comment. Most notable: the multi-instance vault-cleanup fix (the main high-severity finding of Round 1) lacks a regression test.

### What Looks Good
- The `BeginOrContinue`/`owned` + deferred `RollbackTx` + conditional `CommitTx` pattern is implemented correctly, with every query routed through `tx`.
- `ListAgentConnectorCredentialsForAgentConnector` correctly fetches bindings *before* the cascading connector delete and cleans vault secrets after — ordering is right.
- The `requireAgentConnectorInstance` helper replaced 4 duplicated blocks with one call; good DRY win while keeping error codes consistent.
- New error code `connector_not_enabled` plumbed through `AgentConnectorErrCode` → handler → 400 response is cleanly layered.
- Label length validation is belt-and-braces: OpenAPI (`minLength`/`maxLength`), Go (`ErrAgentConnectorInstanceLabelRequired`/`ErrAgentConnectorInstanceLabelTooLong`), and the DB unique index — all consistent at 1–256 runes.

---

---

chiedocursor:
⚠️ **Auto-merge failed.** The `--automerge` flag was set, but the merge could not be completed. Please merge manually.

chiedoclaude:
## Code Review Complete

**PR:** #964 — Phase 2: Backend connector instance CRUD and API routes
**Author:** @chiedocursor
**Rounds completed:** 5/6
**Reason for stopping:** 3 consecutive clean rounds with no new findings or commits

### Outstanding Issues
One low-severity test-coverage gap remains from Round 2:

- No regression test for the multi-instance vault-cleanup fix in `handleDisableAgentConnector` (the Round 1 High-severity finding). A single test that creates two instances with distinct credentials and asserts both `vault.DeleteSecret` calls fire on `delete_credentials=true` would lock this in.
- No API tests for the instance credential routes (`PUT/GET/DELETE .../instances/{id}/credential`).
- No cross-tenant 404 tests for the new instance routes.

None block merge — they're follow-ups to harden the test suite.

### Review History
| 1 | 7 | Initial review — 3 high, 3 medium, 1 low |
| 2 | 1 | Commit `f104b4d` fixed 6 of 7 Round 1 findings; 1 low-severity test-coverage gap flagged |
| 3 | 0 | No new commits — clean |
| 4 | 0 | No new commits — clean |
| 5 | 0 | No new commits — clean; exit on 3 consecutive clean rounds |

### Overall Assessment
Solid, focused PR. Phase 2 cleanly layers instance-aware CRUD on top of the Phase 1 schema/backfill, with legacy routes routed through the default instance for back-compat. Commit `f104b4d` addressed every substantive correctness and validation issue from Round 1 with good engineering: the `BeginOrContinue`/`CommitTx` wrap on `DeleteAgentConnectorInstance`, the `requireAgentConnectorInstance` helper that DRYs four handler blocks into one, the multi-instance vault cleanup in `handleDisableAgentConnector`, and the belt-and-braces label validation across OpenAPI, Go, and the DB unique index all reflect careful execution.

The remaining test gaps are the main follow-up — particularly the one for the multi-instance disable path, since that behavior would be easy to silently regress in a future refactor. Everything else (architecture, validation, error mapping, deprecation markers on legacy routes) looks ready to merge once CI passes and manual verification of the test-plan checkboxes is complete.

---

---

chiedo (approved):

PR #965, opened 2026-04-20
Phase 3: Connector instances dashboard UI (issue #962)

chiedocursor:
Implements **Phase 3 — Frontend UI** from #962: multi-instance connector management on the agent connector config page, plus backend support needed for **Make default** (`PATCH` with `is_default: true`).

### Summary

- New `ConnectorInstancesSection` lists instances as cards with inline rename, per-instance credential dropdown (same patterns as the former single-instance UI), **Make default**, **Disconnect** (clears binding), **Remove instance** (non-default), and **Add another**.
- `ConnectorConfigPage` renders the new section instead of `ConnectorCredentialsSection`.
- `DisableConnectorSection` loads instance + credential bindings to show how many instances will be removed and disconnects all linked OAuth connections on **Remove** when applicable.
- **API/DB**: `SetDefaultAgentConnectorInstance` + extended `PATCH .../instances/{instance_id}` (optional `label` and/or `is_default`). OpenAPI `PatchAgentConnectorInstanceRequest` updated; bundle regenerated.

### Developer

- [x] Frontend hooks and UI; tests (`make test-frontend` passed locally)
- [x] OpenAPI + bundle updated for PATCH body
- [ ] CI `go test` / full repo checks (sandbox cannot run Go module graph; rely on CI)

### Manual Testing

- [ ] Add a second instance, assign distinct OAuth per instance, set default, rename, disconnect, remove non-default instance
- [ ] Disable connector with multiple instances: confirm copy mentions count and OAuth disconnect behavior
- [ ] Single-instance agents: one card, familiar credential row

Ref #962

<div><a href="https://cursor.com/agents/bc-be8b9c07-a693-4045-9901-c2862b939b5d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-be8b9c07-a693-4045-9901-c2862b939b5d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/api/agent_connector_instances.go
+++ b/api/agent_connector_instances.go
@@ -29,7 +29,8 @@ type createAgentConnectorInstanceRequest struct {
 }
 
 type patchAgentConnectorInstanceRequest struct {
-	Label string `json:"label"`
+	Label     *string `json:"label,omitempty"`
+	IsDefault *bool   `json:"is_default,omitempty"`
 }
 
 func init() {
@@ -242,25 +243,59 @@ func handlePatchAgentConnectorInstance(deps *Deps) http.HandlerFunc {
 		if !DecodeJSONOrReject(w, r, &req) {
 			return
 		}
-
-		inst, err := db.RenameAgentConnectorInstance(r.Context(), deps.DB, agentID, userID, connectorID, instanceID, req.Label)
-		if err != nil {
-			if errors.Is(err, db.ErrAgentConnectorInstanceLabelRequired) || errors.Is(err, db.ErrAgentConnectorInstanceLabelTooLong) {
-				RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "label must be 1–256 characters"))
-				return
-			}
-			var instErr *db.AgentConnectorInstanceError
-			if errors.As(err, &instErr) && instErr.Code == db.AgentConnectorInstanceErrDuplicateLabel {
-				RespondError(w, r, http.StatusConflict, Conflict(ErrConstraintViolation, "An instance with this label already exists"))
-				return
-			}
-			log.Printf("[%s] RenameAgentConnectorInstance: %v", TraceID(r.Context()), err)
-			CaptureError(r.Context(), err)
-			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to rename connector instance"))
+		if req.Label == nil && req.IsDefault == nil {
+			RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "Provide label and/or is_default"))
 			return
 		}
+		if req.IsDefault != nil && !*req.IsDefault {
+			RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "is_default may only be set to true"))
+			return
+		}
+
+		ctx := r.Context()
+		var inst *db.AgentConnectorInstance
+
+		if req.Label != nil {
+			var err error
+			inst, err = db.RenameAgentConnectorInstance(ctx, deps.DB, agentID, userID, connectorID, instanceID, *req.Label)
+			if err != nil {
+				if errors.Is(err, db.ErrAgentConnectorInstanceLabelRequired) || errors.Is(err, db.ErrAgentConnectorInstanceLabelTooLong) {
+					RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "label must be 1–256 characters"))
+					return
+				}
+				var instErr *db.AgentConnectorInstanceError
+				if errors.As(err, &instErr) && instErr.Code == db.AgentConnectorInstanceErrDuplicateLabel {
+					RespondError(w, r, http.StatusConflict, Conflict(ErrConstraintViolation, "An instance with this label already exists"))
+					return
+				}
+				log.Printf("[%s] RenameAgentConnectorInstance: %v", TraceID(ctx), err)
+				CaptureError(ctx, err)
+				RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to rename connector instance"))
+				return
+			}
+			if inst == nil {
+				RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+				return
+			}
+		}
+
+		if req.IsDefault != nil && *req.IsDefault {
+			def, err := db.SetDefaultAgentConnectorInstance(ctx, deps.DB, agentID, userID, connectorID, instanceID)
+			if err != nil {
+				log.Printf("[%s] SetDefaultAgentConnectorInstance: %v", TraceID(ctx), err)
+				CaptureError(ctx, err)
+				RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to update default connector instance"))
+				return
+			}
+			if def == nil {
+				RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+				return
+			}
+			inst = def
+		}
+
 		if inst == nil {
-			RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to update connector instance"))
 			return
 		}
 		RespondJSON(w, http.StatusOK, toAgentConnectorInstanceResponse(*inst))
--- a/db/agent_connector_instances.go
+++ b/db/agent_connector_instances.go
@@ -216,6 +216,52 @@ func GetDefaultAgentConnectorInstanceByAgent(ctx context.Context, db DBTX, agent
 	return inst, nil
 }
 
+// SetDefaultAgentConnectorInstance marks the given instance as the default for
+// (agent, connector) and clears is_default on sibling instances.
+func SetDefaultAgentConnectorInstance(ctx context.Context, db DBTX, agentID int64, approverID, connectorID, connectorInstanceID string) (*AgentConnectorInstance, error) {
+	tx, owned, err := BeginOrContinue(ctx, db)
+	if err != nil {
+		return nil, err
+	}
+	if owned {
+		defer RollbackTx(ctx, tx) //nolint:errcheck // commit path clears
+	}
+
+	_, err = tx.Exec(ctx, `
+		UPDATE agent_connectors
+		SET is_default = false
+		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
+		  AND is_default
+		  AND connector_instance_id <> $4::uuid`,
+		agentID, approverID, connectorID, connectorInstanceID,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	row := tx.QueryRow(ctx, `
+		UPDATE agent_connectors
+		SET is_default = true
+		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3 AND connector_instance_id = $4::uuid
+		RETURNING connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at`,
+		agentID, approverID, connectorID, connectorInstanceID,
+	)
+	inst, err := scanAgentConnectorInstance(row)
+	if errors.Is(err, pgx.ErrNoRows) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if owned {
+		if err := CommitTx(ctx, tx); err != nil {
+			return nil, err
+		}
+	}
+	return inst, nil
+}
+
 // RenameAgentConnectorInstance updates the label for an instance.
 func RenameAgentConnectorInstance(ctx context.Context, db DBTX, agentID int64, approverID, connectorID, connectorInstanceID, newLabel string) (*AgentConnectorInstance, error) {
 	label := strings.TrimSpace(newLabel)
--- a/frontend/src/hooks/useConnectorInstanceBindingsSummary.ts
+++ b/frontend/src/hooks/useConnectorInstanceBindingsSummary.ts
@@ -1,0 +1,71 @@
+import { useQuery } from "@tanstack/react-query";
+import { useAuth } from "@/auth/AuthContext";
+import client from "@/api/client";
+
+export interface ConnectorInstanceBindingsSummary {
+  instanceCount: number;
+  oauthConnectionIds: string[];
+  hasApiKeyCredential: boolean;
+}
+
+/**
+ * Loads connector instances and each instance's credential binding for danger-zone
+ * copy and OAuth disconnect-on-remove.
+ */
+export function useConnectorInstanceBindingsSummary(
+  agentId: number,
+  connectorId: string,
+) {
+  const { session } = useAuth();
+  const accessToken = session?.access_token;
+
+  return useQuery({
+    queryKey: ["connector-instance-bindings-summary", agentId, connectorId],
+    queryFn: async (): Promise<ConnectorInstanceBindingsSummary> => {
+      if (!accessToken) throw new Error("Missing access token");
+      const { data: listData, error: listErr } = await client.GET(
+        "/v1/agents/{agent_id}/connectors/{connector_id}/instances",
+        {
+          headers: { Authorization: `Bearer ${accessToken}` },
+          params: { path: { agent_id: agentId, connector_id: connectorId } },
+        },
+      );
+      if (listErr) throw new Error("Failed to list instances");
+      const instances = listData?.data ?? [];
+      const oauthIds: string[] = [];
+      let hasApiKeyCredential = false;
+
+      await Promise.all(
+        instances.map(async (inst) => {
+          const { data: credData, error: credErr } = await client.GET(
+            "/v1/agents/{agent_id}/connectors/{connector_id}/instances/{instance_id}/credential",
+            {
+              headers: { Authorization: `Bearer ${accessToken}` },
+              params: {
+                path: {
+                  agent_id: agentId,
+                  connector_id: connectorId,
+                  instance_id: inst.connector_instance_id,
+                },
+              },
+            },
+          );
+          if (credErr || !credData) return;
+          if (credData.oauth_connection_id) {
+            oauthIds.push(credData.oauth_connection_id);
+          }
+          if (credData.credential_id) {
+            hasApiKeyCredential = true;
+          }
+        }),
+      );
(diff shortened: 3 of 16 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 8 (0 critical, 2 high, 4 medium, 2 low)

Phase 3 is well-executed overall — the hook layer, invalidation graph, and test coverage for the happy paths are solid. The main concerns are around correctness under edge cases: `PATCH /instances/{id}` runs rename + set-default as two separate transactions (partial state possible), `SetDefaultAgentConnectorInstance` has a latent atomicity hazard when used inside an outer transaction, and the danger-zone summary silently undercounts OAuth connections on fetch failure. Everything else is maintainability/UX polish: the 626-line `ConnectorInstancesSection.tsx` should be split, the delete dialog should guard against closing mid-request for consistency with `DisableConnectorSection`, and a couple of derived values deserve `useMemo`.

### What Looks Good
- Clean cache-invalidation graph across the four new mutation hooks and the legacy credential hook — `connector-instance-bindings-summary` is correctly invalidated from every mutation that could change it.
- Good use of pointer/nullable fields on `patchAgentConnectorInstanceRequest` to distinguish omitted from zero-valued updates, including the explicit `is_default: false` rejection.
- Strong test coverage for the new endpoints (create/list/patch set-default/patch duplicate-label/delete) and a solid DB-level test for the standing approval revocation path.
- OpenAPI description is updated with the "at least one of" rule.

---

---

db/agent_connector_instances.go
```diff
@@ -216,6 +216,52 @@ func GetDefaultAgentConnectorInstanceByAgent(ctx context.Context, db DBTX, agent
...
+
+	row := tx.QueryRow(ctx, `
+		UPDATE agent_connectors
+		SET is_default = true
+		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3 AND connector_instance_id = $4::uuid
+		RETURNING connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at`,
+		agentID, approverID, connectorID, connectorInstanceID,
+	)
+	inst, err := scanAgentConnectorInstance(row)
+	if errors.Is(err, pgx.ErrNoRows) {
```
chiedoclaude, line 250: **[High · Correctness]** Latent atomicity hazard when this function is called inside an outer transaction.

The first `UPDATE` on line 230 unconditionally clears `is_default = false` from the current default (a *different* instance from the target). Then the second `UPDATE` tries to set `is_default = true` on the target. If the target doesn't exist, we hit `pgx.ErrNoRows` here and return `nil, nil`.

When `owned == true` (production path from `handlePatchAgentConnectorInstance`), the deferred `RollbackTx` saves us. But when `owned == false` (caller supplied an existing `tx`, e.g., for composing with other writes), we leave the outer tx in an inconsistent state: the previous default has been un-defaulted, no new default was set, and the caller sees `nil, nil` with no signal that they should roll back. The caller may happily commit an (agent, connector) pair with *zero* default instances — violating the invariant that powers `GetDefaultAgentConnectorInstance`.

**Why it matters:** Currently only `handlePatchAgentConnectorInstance` calls this with the pool, so the bug is latent. But the `BeginOrContinue` pattern clearly invites composition, and this footgun will fire the first time someone wraps it.

**Fix:** Do the existence-check-and-swap atomically in a single statement so the "not found" branch makes no writes:

```go
row := tx.QueryRow(ctx, `
    WITH target AS (
        SELECT 1 FROM agent_connectors
        WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
          AND connector_instance_id = $4::uuid
    ),
    cleared AS (
        UPDATE agent_connectors SET is_default = false
        WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
          AND is_default
          AND connector_instance_id <> $4::uuid
          AND EXISTS (SELECT 1 FROM target)
    )
    UPDATE agent_connectors SET is_default = true
    WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
      AND connector_instance_id = $4::uuid
    RETURNING connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at`,
    agentID, approverID, connectorID, connectorInstanceID,
)
```

That way the "clear others" step only fires if the target exists — no rollback required for the negative case.

---

api/agent_connector_instances.go
```diff
@@ -242,25 +243,59 @@ func handlePatchAgentConnectorInstance(deps *Deps) http.HandlerFunc {
...
+				CaptureError(ctx, err)
+				RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to update default connector instance"))
+				return
+			}
+			if def == nil {
+				RespondError(w, r, http.StatusNotFound, NotFound(ErrConnectorInstanceNotFound, "Connector instance not found"))
+				return
+			}
+			inst = def
+		}
```
chiedoclaude, line 295: **[High · Correctness]** `PATCH` with both `label` and `is_default: true` is not atomic — the rename commits in one transaction and the default-swap runs in a second.

If `SetDefaultAgentConnectorInstance` returns a transient error (connection blip, deadlock, ctx cancellation) after a successful rename, the client gets HTTP 500 and will likely retry. Meanwhile the label change has already been committed, so on retry the client may see "label already matches" surprises, or (if the retry body was just `is_default: true`) they'll have a renamed-but-not-defaulted instance and no indication that half the work succeeded.

**Why it matters:** clients usually assume PATCH is atomic on the server. Splitting it into two independent writes means partial state is observable and clients have to be defensive.

**Fix — pick one:**
1. Wrap both calls in a single `BeginOrContinue` transaction at the handler layer so rename + set-default commit or rollback together (cleanest; the db funcs already support an external tx).
2. Reject combined updates with a 400 ("label and is_default cannot be set in the same request") until (1) is in place — low effort, preserves the one-thing-per-call guarantee.

Either is fine; (1) matches what the API surface advertises.

---

frontend/src/hooks/useConnectorInstanceBindingsSummary.ts
```diff
@@ -0,0 +1,71 @@
...
+              params: {
+                path: {
+                  agent_id: agentId,
+                  connector_id: connectorId,
+                  instance_id: inst.connector_instance_id,
+                },
+              },
+            },
+          );
+          if (credErr || !credData) return;
```
chiedoclaude, line 53: **[Medium · Correctness / UX]** Per-instance credential fetch errors are swallowed silently, which lets the danger-zone copy *undercount* OAuth connections.

If any of the parallel `/credential` GETs fails, that instance contributes nothing to `oauthIds` or `hasApiKeyCredential`. The `DisableConnectorSection` then tells the user "disconnect *the* GitHub OAuth connection" or even "permanently delete the saved API key" when in reality there may be another OAuth connection still bound that will quietly survive the remove flow (the subsequent remove code also iterates only this list). The user thinks they've fully removed the connector but haven't.

**Why it matters:** this is a danger-zone UI that users rely on to understand consequences before clicking Remove. Silently missing an OAuth connection is exactly the kind of thing that causes "why is this still granted?" tickets later.

**Fix:** Let a fetch failure propagate so React Query reports an error and the danger zone can render a "we couldn't enumerate bindings — refresh before proceeding" state rather than a wrong number. Minimum viable:

```ts
const results = await Promise.all(
  instances.map(async (inst) => {
    const { data, error } = await client.GET(
      "/v1/agents/{agent_id}/connectors/{connector_id}/instances/{instance_id}/credential",
      { /* ... */ },
    );
    if (error) throw new Error("Failed to load credential binding for an instance");
    return data;
  }),
);
```

Even better: expose a batch endpoint on the backend so this isn't N+1 in the first place — 10+ instances means 11 round-trips just to render the danger-zone header.

---

frontend/src/pages/agents/connectors/ConnectorInstancesSection.tsx
```diff
@@ -0,0 +1,626 @@
...
+export interface ConnectorInstancesSectionProps {
+  agentId: number;
+  connectorId: string;
+  requiredCredentials: RequiredCredential[];
+}
+
+const selectClassName =
+  "border-input bg-background flex h-9 w-full rounded-md border px-3 py-1 text-sm";
+
+export function ConnectorInstancesSection({
```
chiedoclaude, line 65: **[Medium · Maintainability]** This file ends up at 626 lines with three top-level components (`ConnectorInstancesSection`, `InstanceCard`, `InstanceCredentialBinding`), and `CLAUDE.md` is explicit: *"If a file is getting large enough that splitting it would improve maintainability, just go ahead and split it — don't ask first."*

**Why it matters:** this is also a hot file — it composes hooks that every Phase 3 follow-up will touch, and long files in parallel-edited codebases are merge-conflict magnets (per the CLAUDE.md guidance two sections down).

**Fix:** Extract `InstanceCard` → `InstanceCard.tsx` and `InstanceCredentialBinding` → `InstanceCredentialBinding.tsx`, leaving this file focused on the section shell + "Add another" dialog (~220 lines). The components are already pure/self-contained; the split is mechanical.

---

frontend/src/pages/agents/connectors/ConnectorInstancesSection.tsx
```diff
@@ -0,0 +1,626 @@
...
+          connectorId={connectorId}
+          instanceId={instance.connector_instance_id}
+          credentialServiceIds={credentialServiceIds}
+          credentials={credentials}
+          connections={connections}
+          anyLoading={anyLoading}
+        />
+      </div>
+
+      <Dialog open={deleteOpen} onOpenChange={setDeleteOpen}>
```
chiedoclaude, line 433: **[Medium · UX / Inconsistency]** The delete confirmation dialog can be dismissed (Esc / click-outside) while `deleteInstance` is in flight, which cancels the modal but *doesn't* cancel the mutation. The user is then left with a toast but no confirmation dialog and a mutated list — and the Cancel button isn't disabled either (only the destructive button is).

Compare this to `DisableConnectorSection.tsx` line 235, which guards correctly:
```tsx
<Dialog open={removeConfirmOpen} onOpenChange={(open) => !isRemoving && setRemoveConfirmOpen(open)}>
```

**Why it matters:** destructive-action dialogs should behave consistently across the product, and leaving mid-request gives users the wrong mental model about whether the delete was aborted.

**Fix:**
```tsx
<Dialog open={deleteOpen} onOpenChange={(open) => !deleting && setDeleteOpen(open)}>
```
and add `disabled={deleting}` to the Cancel button for symmetry with the Remove button.

---

frontend/src/pages/agents/connectors/ConnectorInstancesSection.tsx
```diff
@@ -0,0 +1,626 @@
...
+
+  const matchingProvider = providers.find((p) => p.id === connectorId);
+  const hasImplicitOAuth = !hasExplicitOAuth && !!matchingProvider;
+  const hasOAuth = hasExplicitOAuth || hasImplicitOAuth;
+
+  const sorted = [...requiredCredentials].sort((a, b) => {
+    if (a.auth_type === "oauth2" && b.auth_type !== "oauth2") return -1;
+    if (a.auth_type !== "oauth2" && b.auth_type === "oauth2") return 1;
+    return 0;
+  });
```
chiedoclaude, line 117: **[Low · Performance]** Both `storedByService` (lines 102–107) and `sorted` (lines 113–117) are rebuilt on every render of `ConnectorInstancesSection`. Each `InstanceCard` then receives `credentials` / `connections` unchanged, so adding a new instance or toggling one edit field triggers a full re-sort and re-bucket across all credentials every render.

**Why it matters:** on agents with many credentials or OAuth connections, this isn't free, and the `storedByService` map isn't even used by the rendered tree here — it's only handed to `ManageCredentialsDialog`, which means every render also throws away a reference that Dialog's downstream `useMemo`s depend on.

**Fix:** wrap both in `useMemo` keyed on `credentials` / `requiredCredentials`:
```tsx
const storedByService = useMemo(() => {
  const m = new Map<string, CredentialSummary[]>();
  for (const cred of credentials) {
    const list = m.get(cred.service) ?? [];
    list.push(cred);
    m.set(cred.service, list);
  }
  return m;
}, [credentials]);

const sorted = useMemo(
  () => [...requiredCredentials].sort(...),
  [requiredCredentials],
);
```

---

frontend/src/hooks/useConnectorInstanceBindingsSummary.ts
```diff
@@ -0,0 +1,71 @@
...
+          if (credData.oauth_connection_id) {
+            oauthIds.push(credData.oauth_connection_id);
+          }
+          if (credData.credential_id) {
+            hasApiKeyCredential = true;
+          }
+        }),
+      );
```
chiedoclaude, line 61: **[Medium · Performance]** This is a true N+1: one `GET /instances` followed by one `GET /instances/{id}/credential` per instance, all fired from the browser. For an agent with, say, 5 Slack workspaces this is 6 round-trips just to render the "Disable connector" section — and the query runs on every `DisableConnectorSection` mount.

**Why it matters:** the danger zone is far from the critical path, but it's visible on every connector config page load, not only when the user clicks Remove. Under the `hasApiKeyCredential || oauthProvider` branch it's also blocking the Remove button's copy.

**Fix options (in order of effort):**
1. Add a backend endpoint that returns instance + binding summary in one call (ideal, gives you atomic consistency too).
2. Make this hook lazy — only enable the query once the user opens the remove dialog (`enabled: removeConfirmOpen`).
3. At minimum memoize / cache-share the per-instance credential fetches with `useAgentConnectorInstanceCredential` so opening the per-card select afterwards doesn't re-fetch.

Option 2 is a one-line `enabled:` change and buys most of the win if (1) is too much for this PR.

---

api/agent_connector_instances.go
```diff
@@ -242,25 +243,59 @@ func handlePatchAgentConnectorInstance(deps *Deps) http.HandlerFunc {
...
 			return
 		}
+		if req.Label == nil && req.IsDefault == nil {
+			RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "Provide label and/or is_default"))
+			return
+		}
+		if req.IsDefault != nil && !*req.IsDefault {
+			RespondError(w, r, http.StatusBadRequest, BadRequest(ErrInvalidRequest, "is_default may only be set to true"))
+			return
+		}
```
chiedoclaude, line 253: **[Low · Testing]** The new validation branches at lines 246–253 aren't covered. It'd be cheap to add a couple of assertions to `api/agent_connector_instances_test.go`:

- `PATCH` with empty body `{}` → expect 400 with "Provide label and/or is_default".
- `PATCH` with `{"is_default": false}` → expect 400 with "is_default may only be set to true".
- `PATCH` with `{"label": "foo", "is_default": true}` on a non-default instance → expect 200, label changed, now default. (This also nails down the combined-update semantics that the OpenAPI schema advertises but currently isn't exercised.)

The third one especially will pay for itself once the partial-state concern on the other review comment is addressed.

---

chiedoclaude (commented):
## Review Round 1/6 — inline findings

**Findings:** 8 (0 critical, 2 high, 4 medium, 2 low)

The summary comment was already posted above; this review attaches the inline comments to the specific lines. See the "What Looks Good" section in the previous summary — the happy-path design is solid, these are all sharp-edges and polish items.

Highlights:
- **High:** `SetDefaultAgentConnectorInstance` has a latent atomicity hazard when composed inside an outer transaction.
- **High:** `PATCH` rename + set-default execute in two separate transactions; partial state is observable.
- **Medium:** `useConnectorInstanceBindingsSummary` silently undercounts OAuth bindings on fetch failure in a danger-zone UI.
- **Medium:** `ConnectorInstancesSection.tsx` (626 lines) should be split per CLAUDE.md guidance.
- **Medium:** delete dialog lets users dismiss mid-request, unlike the sibling `DisableConnectorSection`.
- **Medium:** N+1 pattern in `useConnectorInstanceBindingsSummary` fires on every page load.
- **Low:** memoize derived values in `ConnectorInstancesSection`.
- **Low:** tests don't cover new PATCH validation branches.

---

---

chiedoclaude:
## Code Review Complete

**PR:** #965 — Phase 3: Connector instances dashboard UI (issue #962)
**Author:** @chiedocursor
**Rounds completed:** 2/6
**Reason for stopping:** PR merged (merged at 2026-04-20 23:49 UTC during Round 3's wait window)

### Outstanding Issues
The PR merged before the inline findings from Round 1 could be addressed. Flagging the two highest-severity items so they can be tracked as follow-up work:

- **High · Correctness — `db/agent_connector_instances.go`:** `SetDefaultAgentConnectorInstance` clears the current default before verifying the target exists. When composed inside an outer transaction (`owned == false`), the "not found" branch returns `nil, nil` while leaving the outer tx with no default row for that (agent, connector). The single-statement CTE fix in the inline comment avoids the hazard entirely.
- **High · Correctness — `api/agent_connector_instances.go`:** `PATCH` with both `label` and `is_default: true` runs as two independent transactions; a transient failure on the second call yields a committed rename + 500 response. Recommend wrapping both in a shared `BeginOrContinue` at the handler layer, or rejecting combined updates until then.

Other items (danger-zone summary undercount, N+1 bindings fetch, 626-line `ConnectorInstancesSection.tsx`, delete-dialog mid-request dismissal, memoization, test coverage for PATCH validation branches) are documented inline and worth picking up in a follow-up PR.

### Review History
| 1 | 8 | Initial review — 2 high, 4 medium, 2 low |
| 2 | 0 | Clean round — no new commits, no new findings |

### Overall Assessment
Phase 3 is a competent, well-tested feature PR. The hook layer is clean, cache invalidation covers all mutation paths, backend tests cover create/list/patch/delete happy paths, and the OpenAPI schema is up-to-date. The concerns are at the edges: two correctness issues around transactional atomicity (high severity but narrow blast radius today — only one caller in each case) and a handful of maintainability/UX polish items.

Given the PR has merged, I'd recommend opening a small follow-up issue for the two High findings so the `SetDefault` CTE refactor and the PATCH transaction wrapping don't get lost. The Medium/Low items can be rolled into whichever branch touches those files next.

---

---

PR #967, opened 2026-04-21
Phase 4: Tool-call connector instance routing

### LINKED ISSUE #962: Support multiple instances of the same connector type per agent

chiedocursor:
## Summary

Implements GitHub issue #962 Phase 4: agents can target a specific connector instance via `action.parameters.connector_instance` (label or UUID). The server strips it before schema validation, freezes `_connector_instance_id` / `_connector_instance_label` on the stored action JSON, and routes credentials + standing-approval matching accordingly.

Also extends capabilities with per-instance rows (`instances[]`), threads instance scope through execution paths, enriches approval-request audit metadata, and updates web + mobile approval UIs to show `Connector (Label)` when a frozen label exists.

## Test plan

- [x] `cd frontend && npx vitest run src/pages/dashboard/approvalConnectorLabel.test.ts`
- [x] `cd mobile && npm test -- --ci src/screens/approvals/__tests__/approvalUtils.test.ts`
- [ ] Backend: `make test-backend` — **not run in this environment** (local Go version mismatch vs `go.mod`; CI should run full suite)

Closes #962

<div><a href="https://cursor.com/agents/bc-9e6ab8f7-c0d0-4bbe-8fc4-be0ffa6e989a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9e6ab8f7-c0d0-4bbe-8fc4-be0ffa6e989a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/api/agent_approvals.go
+++ b/api/agent_approvals.go
@@ -18,13 +18,13 @@ import (
 // ── Request/response types ─────────────────────────────────────────────────
 
 type agentRequestApprovalRequest struct {
-	RequestID       string                 `json:"request_id" validate:"required"`
-	Action          json.RawMessage        `json:"action" validate:"required"`
-	Context         json.RawMessage        `json:"context" validate:"required"`
-	ExpiresIn       *int                   `json:"expires_in,omitempty" validate:"omitempty,gte=60,lte=86400"`
+	RequestID       string                  `json:"request_id" validate:"required"`
+	Action          json.RawMessage         `json:"action" validate:"required"`
+	Context         json.RawMessage         `json:"context" validate:"required"`
+	ExpiresIn       *int                    `json:"expires_in,omitempty" validate:"omitempty,gte=60,lte=86400"`
 	Configuration   *agentApprovalConfigRef `json:"configuration,omitempty"`
-	PaymentMethodID string                 `json:"payment_method_id,omitempty"`
-	AmountCents     *int                   `json:"amount_cents,omitempty"`
+	PaymentMethodID string                  `json:"payment_method_id,omitempty"`
+	AmountCents     *int                    `json:"amount_cents,omitempty"`
 }
 
 type agentApprovalConfigRef struct {
@@ -204,6 +204,23 @@ func handleAgentRequestApproval(deps *Deps) http.HandlerFunc {
 			}
 		}
 
+		// Multi-instance routing: resolve optional parameters.connector_instance, freeze UUID+label on action, strip from parameters.
+		connectorInstanceID, err := applyConnectorInstanceToAction(r.Context(), deps.DB, agent, actionType, actionObj)
+		if err != nil {
+			var ciErr *connectorInstanceResolutionError
+			if errors.As(err, &ciErr) {
+				RespondError(w, r, ciErr.status, ciErr.resp)
+				return
+			}
+			log.Printf("[%s] applyConnectorInstanceToAction: %v", TraceID(r.Context()), err)
+			CaptureError(r.Context(), err)
+			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to process action"))
+			return
+		}
+		if updated, mErr := json.Marshal(actionObj); mErr == nil {
+			req.Action = updated
+		}
+
 		// Validate action parameters against the connector's parameters_schema.
 		// Runs after alias normalization so canonical keys are checked.
 		// Fail-open: skipped if the action has no schema.
@@ -233,7 +250,7 @@ func handleAgentRequestApproval(deps *Deps) http.HandlerFunc {
 		// approval matches this agent + action type + parameters. If so,
 		// execute immediately and return the result — no human review needed.
 		pp := &paymentParams{PaymentMethodID: req.PaymentMethodID, AmountCents: req.AmountCents}
-		if handled := tryStandingApprovalAutoApprove(w, r, deps, agent, actionType, actionParams, req.RequestID, pp); handled {
+		if handled := tryStandingApprovalAutoApprove(w, r, deps, agent, actionType, actionParams, req.RequestID, pp, connectorInstanceID); handled {
 			return
 		}
 
@@ -272,7 +289,7 @@ func handleAgentRequestApproval(deps *Deps) http.HandlerFunc {
 				if conn, ok := deps.Connectors.Get(cid[0]); ok {
 					if resolver, ok := conn.(connectors.ResourceDetailResolver); ok {
 						resolveCtx, resolveCancel := context.WithTimeout(r.Context(), 5*time.Second)
-						details, resolveErr := resolver.ResolveResourceDetails(resolveCtx, actionType, actionParams, resolveCredentialsForResolver(resolveCtx, deps, agent.AgentID, agent.ApproverID, actionType, cid[0]))
+						details, resolveErr := resolver.ResolveResourceDetails(resolveCtx, actionType, actionParams, resolveCredentialsForResolver(resolveCtx, deps, agent.AgentID, agent.ApproverID, actionType, cid[0], connectorInstanceID))
 						resolveCancel()
 						if resolveErr != nil {
 							log.Printf("[%s] ResolveResourceDetails: %v", TraceID(r.Context()), resolveErr)
@@ -432,12 +449,12 @@ var handleAgentApprovalError = handleApprovalError
 // ResourceDetailResolver call. It mirrors the credential resolution logic from
 // connector execution but returns empty credentials on any failure (since
 // resource detail resolution is non-fatal).
-func resolveCredentialsForResolver(ctx context.Context, deps *Deps, agentID int64, userID, actionType, connectorID string) connectors.Credentials {
+func resolveCredentialsForResolver(ctx context.Context, deps *Deps, agentID int64, userID, actionType, connectorID, connectorInstanceID string) connectors.Credentials {
 	reqCreds, err := db.GetRequiredCredentialsByActionType(ctx, deps.DB, actionType)
 	if err != nil || len(reqCreds) == 0 {
 		return connectors.NewCredentials(nil)
 	}
-	creds, err := resolveCredentialsWithFallback(ctx, deps, agentID, userID, actionType, connectorID, reqCreds)
+	creds, err := resolveCredentialsWithFallback(ctx, deps, agentID, userID, actionType, connectorID, connectorInstanceID, reqCreds)
 	if err != nil {
 		return connectors.NewCredentials(nil)
 	}
@@ -456,7 +473,7 @@ func emitApprovalRequestAuditEvent(ctx context.Context, d db.DBTX, userID string
 		SourceID:    appr.ApprovalID,
 		SourceType:  "approval",
 		AgentMeta:   agentMeta,
-		Action:      redactActionToType(appr.Action),
+		Action:      redactActionToTypeWithConnectorInstance(appr.Action),
 		ConnectorID: connectorIDFromActionType(actionTypeFromJSON(appr.Action)),
 	}, true)
 }
--- a/api/agent_approvals_instance.go
+++ b/api/agent_approvals_instance.go
@@ -1,0 +1,112 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"strings"
+
+	"github.com/supersuit-tech/permission-slip/db"
+)
+
+// connectorInstanceResolutionError carries an HTTP error for applyConnectorInstanceToAction.
+type connectorInstanceResolutionError struct {
+	status int
+	resp   ErrorResponse
+}
+
+func (e *connectorInstanceResolutionError) Error() string {
+	return e.resp.Error.Message
+}
+
+// applyConnectorInstanceToAction removes parameters.connector_instance, resolves the target
+// instance for the action's connector, and sets _connector_instance_id and _connector_instance_label
+// on the action object. Returns the resolved instance UUID for credential routing (empty if the
+// connector has no enabled instances on the agent — downstream "no credential" handling applies).
+func applyConnectorInstanceToAction(ctx context.Context, d db.DBTX, agent *db.Agent, actionType string, actionObj map[string]json.RawMessage) (connectorInstanceID string, err error) {
+	parts := strings.SplitN(actionType, ".", 2)
+	if len(parts) != 2 {
+		return "", nil
+	}
+	connectorID := parts[0]
+
+	var paramsObj map[string]json.RawMessage
+	if raw, ok := actionObj["parameters"]; ok && len(raw) > 0 && string(raw) != "null" {
+		if err := json.Unmarshal(raw, &paramsObj); err != nil {
+			return "", err
+		}
+	}
+
+	var selector string
+	if paramsObj != nil {
+		if raw, ok := paramsObj["connector_instance"]; ok {
+			_ = json.Unmarshal(raw, &selector)
+			selector = strings.TrimSpace(selector)
+			delete(paramsObj, "connector_instance")
+			if len(paramsObj) == 0 {
+				delete(actionObj, "parameters")
+			} else {
+				b, mErr := json.Marshal(paramsObj)
+				if mErr != nil {
+					return "", mErr
+				}
+				actionObj["parameters"] = b
+			}
+		}
+	}
+
+	instances, err := db.ListAgentConnectorInstances(ctx, d, agent.AgentID, agent.ApproverID, connectorID)
+	if err != nil {
+		return "", err
+	}
+	if len(instances) == 0 {
+		return "", nil
+	}
+
+	var inst *db.AgentConnectorInstance
+	if len(instances) == 1 {
+		only := &instances[0]
+		if selector == "" {
+			inst = only
+		} else {
+			resolved, err := db.ResolveAgentConnectorInstance(ctx, d, agent.AgentID, agent.ApproverID, connectorID, selector)
+			if err != nil {
+				return "", err
+			}
+			if resolved == nil || resolved.ConnectorInstanceID != only.ConnectorInstanceID {
+				return "", &connectorInstanceResolutionError{
+					status: http.StatusBadRequest,
+					resp:   BadRequest(ErrConnectorInstanceNotFound, "no connector instance matches the given connector_instance"),
+				}
+			}
+			inst = resolved
+		}
+	} else {
+		if selector == "" {
+			labels := make([]string, 0, len(instances))
+			for _, i := range instances {
+				labels = append(labels, i.Label)
+			}
+			resp := BadRequest(ErrConnectorInstanceRequired, "connector_instance is required when multiple instances exist")
+			resp.Error.Details = map[string]any{"available_instances": labels}
+			return "", &connectorInstanceResolutionError{status: http.StatusBadRequest, resp: resp}
+		}
+		resolved, err := db.ResolveAgentConnectorInstance(ctx, d, agent.AgentID, agent.ApproverID, connectorID, selector)
+		if err != nil {
+			return "", err
+		}
+		if resolved == nil {
+			return "", &connectorInstanceResolutionError{
+				status: http.StatusBadRequest,
+				resp:   BadRequest(ErrConnectorInstanceNotFound, "no connector instance matches the given connector_instance"),
+			}
+		}
+		inst = resolved
+	}
+
+	idRaw, _ := json.Marshal(inst.ConnectorInstanceID)
+	actionObj["_connector_instance_id"] = idRaw
+	labelRaw, _ := json.Marshal(inst.Label)
+	actionObj["_connector_instance_label"] = labelRaw
+	return inst.ConnectorInstanceID, nil
+}
--- a/api/approvals.go
+++ b/api/approvals.go
@@ -251,7 +251,17 @@ func executeApprovalAction(ctx context.Context, deps *Deps, userID string, appr 
 		_ = json.Unmarshal(appr.Action, &actionObj)
 	}
 
-	result, execErr := executeConnectorAction(ctx, deps, appr.AgentID, userID, actionType, actionObj.Parameters, nil)
+	var connectorInstanceID string
+	if len(appr.Action) > 0 {
+		var raw map[string]json.RawMessage
+		if json.Unmarshal(appr.Action, &raw) == nil {
+			if v, ok := raw["_connector_instance_id"]; ok {
+				_ = json.Unmarshal(v, &connectorInstanceID)
+			}
+		}
+	}
+
+	result, execErr := executeConnectorAction(ctx, deps, appr.AgentID, userID, actionType, actionObj.Parameters, nil, connectorInstanceID)
 
 	var execStatus string
 	var resultJSON json.RawMessage
--- a/api/agent_approvals_test.go
+++ b/api/agent_approvals_test.go
@@ -1025,3 +1025,117 @@ func TestAgentRequestApproval_RequestValidatorAllowsValidChannel(t *testing.T) {
 		t.Fatalf("expected 200 for valid channel ID, got %d: %s", w.Code, w.Body.String())
 	}
 }
+
+func TestAgentRequestApproval_MultiInstance_RequiresConnectorInstance(t *testing.T) {
+	t.Parallel()
+	tx := testhelper.SetupTestDB(t)
+	uid := testhelper.GenerateUID(t)
+	testhelper.InsertUser(t, tx, uid, "u_"+uid[:8])
+
+	pubKeySSH, privKey, err := GenerateEd25519OpenSSHKey()
+	if err != nil {
+		t.Fatalf("generate key: %v", err)
+	}
+	agentID := testhelper.InsertAgentWithPublicKey(t, tx, uid, "registered", pubKeySSH)
+
+	connID := "mi_test_conn"
+	testhelper.InsertConnector(t, tx, connID)
+	testhelper.InsertConnectorAction(t, tx, connID, connID+".ping", "Ping")
+	testhelper.InsertAgentConnector(t, tx, agentID, uid, connID)
+	if _, err := db.CreateAgentConnectorInstance(t.Context(), tx, db.CreateAgentConnectorInstanceParams{
+		AgentID: agentID, ApproverID: uid, ConnectorID: connID, Label: "Beta",
+	}); err != nil {
+		t.Fatalf("second instance: %v", err)
+	}
+
+	reg := connectors.NewRegistry()
+	reg.Register(newTestStubConnector(connID, connID+".ping"))
+	deps := &Deps{DB: tx, SupabaseJWTSecret: testJWTSecret, Connectors: reg}
+	router := NewRouter(deps)
+
+	reqBody := `{"request_id":"req_mi_need_inst","action":{"type":"` + connID + `.ping","parameters":{}},"context":{"description":"x"}}`
+	r := signedJSONRequest(t, http.MethodPost, "/approvals/request", reqBody, privKey, agentID)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, r)
+
+	if w.Code != http.StatusBadRequest {
+		t.Fatalf("expected 400, got %d: %s", w.Code, w.Body.String())
+	}
+	var errResp ErrorResponse
+	if err := json.Unmarshal(w.Body.Bytes(), &errResp); err != nil {
+		t.Fatalf("unmarshal: %v", err)
+	}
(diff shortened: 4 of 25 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

## 🔧 CI / audit remediation

### Remediation round 1

- **ci.yml** — conclusion `failure` — api/approval_flow_multi_instance_test.go used http.MethodPost and http.StatusOK without importing net/http; added import net/http. (`cf6a74d9`)

### Remediation round 2

- **ci.yml** — conclusion `failure` — Tests failed: duplicate credentials (user_id,service,label) when two vault credentials shared service without labels; added InsertCredentialWithVaultSecretIDAndLabel. Multi-instance approval test: parameters key omitted when empty after stripping connector_instance — assert only when parameters present. (`4337602a`)

api/agent_approvals_instance.go
```diff
@@ -0,0 +1,112 @@
...
+	if raw, ok := actionObj["parameters"]; ok && len(raw) > 0 && string(raw) != "null" {
+		if err := json.Unmarshal(raw, &paramsObj); err != nil {
+			return "", err
+		}
+	}
+
+	var selector string
+	if paramsObj != nil {
+		if raw, ok := paramsObj["connector_instance"]; ok {
+			_ = json.Unmarshal(raw, &selector)
```
chiedoclaude, line 43: **[Medium · Correctness]** The error from `json.Unmarshal(raw, &selector)` is discarded, so any non-string value for `connector_instance` (number, bool, null, array, object) silently becomes an empty selector.

**Why:** In the single-instance case the agent is then quietly routed to the only instance; in the multi-instance case the agent gets a `connector_instance_required` error *even though they provided the field*. Both are confusing — the agent sees a misleading error (or unexpected behavior) instead of a clear "`connector_instance` must be a string" message.

**Fix:** Type-check and return a validation error when the value is present but not a JSON string:

```go
if raw, ok := paramsObj["connector_instance"]; ok {
    if err := json.Unmarshal(raw, &selector); err != nil {
        return "", &connectorInstanceResolutionError{
            status: http.StatusBadRequest,
            resp:   BadRequest(ErrInvalidRequest, "connector_instance must be a string (instance label or UUID)"),
        }
    }
    selector = strings.TrimSpace(selector)
    delete(paramsObj, "connector_instance")
    // …rest unchanged
}
```

api/agent_approvals.go
```diff
@@ -204,6 +204,23 @@ func handleAgentRequestApproval(deps *Deps) http.HandlerFunc {
...
+				return
+			}
+			log.Printf("[%s] applyConnectorInstanceToAction: %v", TraceID(r.Context()), err)
+			CaptureError(r.Context(), err)
+			RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to process action"))
+			return
+		}
+		if updated, mErr := json.Marshal(actionObj); mErr == nil {
+			req.Action = updated
+		}
```
chiedoclaude, line 222: **[Medium · Reliability]** `json.Marshal(actionObj)` error is silently swallowed here. If marshal fails after `applyConnectorInstanceToAction` has frozen `_connector_instance_id` and `_connector_instance_label` on the map, `req.Action` keeps the *original* action JSON — so the approval is stored without the frozen fields and without the stripped `connector_instance`, defeating the whole routing contract.

**Why:** Later, `executeApprovalAction` reads `_connector_instance_id` from the stored action and falls back to the default-instance credential when it's missing — so a silent marshal failure here means the wrong credential could be used at execution time, with no log trail pointing at the cause.

**Fix:** Treat the marshal error as a 500 (like `applyConnectorInstanceToAction`'s own errors):

```go
updated, mErr := json.Marshal(actionObj)
if mErr != nil {
    log.Printf("[%s] re-marshal action after connector_instance apply: %v", TraceID(r.Context()), mErr)
    CaptureError(r.Context(), mErr)
    RespondError(w, r, http.StatusInternalServerError, InternalError("Failed to process action"))
    return
}
req.Action = updated
```

In practice `json.Marshal` on a `map[string]json.RawMessage` rarely fails, but the cost of defending against it is trivial and the cost of a silent miscredential at runtime is not.

api/approvals.go
```diff
@@ -251,7 +251,17 @@ func executeApprovalAction(ctx context.Context, deps *Deps, userID string, appr
...
-	result, execErr := executeConnectorAction(ctx, deps, appr.AgentID, userID, actionType, actionObj.Parameters, nil)
+	var connectorInstanceID string
+	if len(appr.Action) > 0 {
+		var raw map[string]json.RawMessage
+		if json.Unmarshal(appr.Action, &raw) == nil {
+			if v, ok := raw["_connector_instance_id"]; ok {
+				_ = json.Unmarshal(v, &connectorInstanceID)
+			}
+		}
+	}
```
chiedoclaude, line 262: **[Low · Maintainability]** `appr.Action` is unmarshaled twice — once into the `actionObj` struct above (for `Parameters`), and again here into a `map[string]json.RawMessage` just to read `_connector_instance_id`.

**Why:** Two parses of the same JSON is wasteful and the split logic is easy to drift — e.g., a future reviewer adding another frozen field (`_connector_instance_label`) will naturally add it next to `_connector_instance_id` and miss that `actionObj` is a separate decode.

**Fix:** Parse once into a map and read both `parameters` and `_connector_instance_id` from it:

```go
var raw map[string]json.RawMessage
if len(appr.Action) > 0 {
    _ = json.Unmarshal(appr.Action, &raw)
}
var parameters json.RawMessage
if v, ok := raw["parameters"]; ok {
    parameters = v
}
var connectorInstanceID string
if v, ok := raw["_connector_instance_id"]; ok {
    _ = json.Unmarshal(v, &connectorInstanceID)
}

result, execErr := executeConnectorAction(ctx, deps, appr.AgentID, userID, actionType, parameters, nil, connectorInstanceID)
```

api/agent_approvals_instance.go
```diff
@@ -0,0 +1,112 @@
...
+				resp:   BadRequest(ErrConnectorInstanceNotFound, "no connector instance matches the given connector_instance"),
+			}
+		}
+		inst = resolved
+	}
+
+	idRaw, _ := json.Marshal(inst.ConnectorInstanceID)
+	actionObj["_connector_instance_id"] = idRaw
+	labelRaw, _ := json.Marshal(inst.Label)
+	actionObj["_connector_instance_label"] = labelRaw
```
chiedoclaude, line 110: **[Low · Maintainability]** `inst.Label` is marshaled unconditionally, so if an instance somehow ends up with an empty label (extension/migration edge cases, future code paths that bypass `CreateAgentConnectorInstance`'s required-label check), you'll freeze `"_connector_instance_label": ""` into the stored action and the audit event.

**Why:** Both `redactActionToTypeWithConnectorInstance` and the mobile/web label readers already skip empty strings, so the empty value is never rendered — which means it's pure noise in stored JSON and audit output.

**Fix:** Only set the label field when it's non-empty:

```go
idRaw, _ := json.Marshal(inst.ConnectorInstanceID)
actionObj["_connector_instance_id"] = idRaw
if inst.Label != "" {
    labelRaw, _ := json.Marshal(inst.Label)
    actionObj["_connector_instance_label"] = labelRaw
}
return inst.ConnectorInstanceID, nil
```

Today the DB-level `CreateAgentConnectorInstance` already rejects empty labels, so this is defensive — but it makes the invariant self-documenting.

api/agent_approvals_instance.go
```diff
@@ -0,0 +1,112 @@
...
+			if err != nil {
+				return "", err
+			}
+			if resolved == nil || resolved.ConnectorInstanceID != only.ConnectorInstanceID {
+				return "", &connectorInstanceResolutionError{
+					status: http.StatusBadRequest,
+					resp:   BadRequest(ErrConnectorInstanceNotFound, "no connector instance matches the given connector_instance"),
+				}
+			}
+			inst = resolved
```
chiedoclaude, line 82: **[Low · Simplicity]** In the `len(instances) == 1` + selector path, you already have `only` in hand — going back through `ResolveAgentConnectorInstance` (which re-queries the DB by UUID or label) just to compare the returned ID against `only.ConnectorInstanceID` adds a round-trip that can only produce `only` or `nil`.

**Why:** It's an avoidable DB hit on the hot path (every single-instance approval request with a selector), and the extra `!=` check is dead code once you realize the function can't return any other instance.

**Fix:** Compare directly against the one instance you have:

```go
if len(instances) == 1 {
    only := &instances[0]
    if selector == "" || selector == only.ConnectorInstanceID || selector == only.Label {
        inst = only
    } else {
        return "", &connectorInstanceResolutionError{
            status: http.StatusBadRequest,
            resp:   BadRequest(ErrConnectorInstanceNotFound, "no connector instance matches the given connector_instance"),
        }
    }
}
```

(Only keep the `uuid.Parse` tolerance if callers might pass a non-canonical UUID form; `ResolveAgentConnectorInstance`'s behavior suggests callers already pass canonical form.)

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 5 (0 critical, 0 high, 2 medium, 3 low)

The overall design is sound: the `parameters.connector_instance` selector is resolved once at approval-request time, the UUID+label are frozen on the stored action JSON, and every downstream execution path (standing-approval auto-approve, pending-approval approve, user-initiated SA execution, resource-detail resolver) now threads the instance ID through credential resolution. Test coverage is thorough — the `TestApprovalFlow_MultiInstance_UsesCorrectCredentialOnApprove` end-to-end test in particular gives good confidence that the right credential actually reaches the connector at execution time.

The findings I posted inline are around robustness: the selector parsing silently coerces non-string inputs, the re-marshal error is swallowed after freezing fields, and a couple of small maintainability items. Nothing blocking.

### Additional observations (non-inline)

- **Audit-event consistency:** `api/approvals.go:496` still calls `redactActionToType` for the `approval.approved` / `approval.denied` / `approval.cancelled` events, while `approval.requested` now uses `redactActionToTypeWithConnectorInstance`. For admins investigating which instance actually ran, it'd be useful to switch the approve/deny/cancel paths over to the new helper too so the frozen instance ID/label appears on every approval-lifecycle event, not just request. That line isn't in the diff so I couldn't inline this. Severity: **Low**.
- **Web/mobile helper duplication:** `connectorInstanceLabelFromAction` lives in `mobile/src/screens/approvals/approvalUtils.ts`, while the web (`ReviewApprovalDialog.tsx`) reinvents the same extraction inline with a verbose `typeof … in … as …` ladder. If these ever diverge (e.g., supporting a localized label field), you'll fix it in two places. Not blocking, but consider extracting a shared helper. Severity: **Low**.
- **Capabilities query shape:** The per-connector `BOOL_OR(NOT EXISTS(...))` in `db/capabilities.go` correlates only on `ac.agent_id` / `c.id`, not on `ac.connector_instance_id`, so BOOL_OR evaluates the same predicate for every row in the group. That's fine (it's just a wrapper needed by GROUP BY), but the code will read more clearly if it uses a simple `MAX(CASE…) = 1` or drops the redundant OR. Severity: **Low** (cosmetic).
- **CI / backend tests:** PR body notes that backend tests weren't run locally due to the Go toolchain mismatch. CI needs to be green before merging — the changes cross a lot of hot paths (credential resolution, standing-approval matching, audit events).

### What Looks Good

- Excellent test coverage: multi-instance required, label freeze, stripping, per-instance execution, and a full request→approve→execute E2E with distinct credentials.
- Clean separation: the new `applyConnectorInstanceToAction` lives in its own file (`api/agent_approvals_instance.go`), keeping the main handler readable.
- The `_connector_instance_*` frozen fields are a nice design — they decouple approval storage from runtime agent-connector state, so deleting/relabeling an instance after approval doesn't corrupt historical records.
- OpenAPI updates are in sync with the code (`ApprovalSummary.action.allOf` extension, `ConnectorInstanceCapability`, `StandingApprovalCapability.connector_instance_id`, plus the long-form `Action` description explaining the selector behavior).
- `resolveCredentialsForConnectorInstance` cleanly becomes a one-liner delegating to `resolveCredentialsWithFallback` — nice DRY win.

---

chiedoclaude:
## Code Review Complete

**PR:** #967 — Phase 4: Tool-call connector instance routing
**Author:** @chiedocursor
**Rounds completed:** 5/6
**Reason for stopping:** PR merged

### Outstanding Issues

5 findings from Round 1 remained open at merge time (none critical/high). For follow-up consideration:

1. **[Medium · Correctness]** `api/agent_approvals_instance.go:43` — `connector_instance` selector silently coerces non-string values to empty, which then produces misleading "required" errors or silent default-instance routing. Add a type check and return a validation error.
2. **[Medium · Reliability]** `api/agent_approvals.go:221` — `json.Marshal` error on the re-serialized action is swallowed; a silent failure here would store the action without the frozen `_connector_instance_id`/`_connector_instance_label`, and the later execute path would fall back to the default-instance credential with no log trail. Treat it as a 500 and log.
3. **[Low · Maintainability]** `api/approvals.go:247–262` — `appr.Action` is unmarshaled twice (once for parameters, once for the instance ID). Consolidate into a single `map[string]json.RawMessage` decode.
4. **[Low · Maintainability]** `api/agent_approvals_instance.go:109–110` — `_connector_instance_label` is frozen even when empty; guard with `if inst.Label != ""`.
5. **[Low · Simplicity]** `api/agent_approvals_instance.go:72–82` — Single-instance + selector path does an unnecessary DB round-trip via `ResolveAgentConnectorInstance`; compare directly against `only.ConnectorInstanceID` / `only.Label`.

Plus three review-body items (audit-event consistency between `approval.requested` and `approval.approved/denied/cancelled`, web/mobile `connectorInstanceLabelFromAction` duplication, cosmetic simplification of the `BOOL_OR` in the capabilities query).

### Review History

| 1 | 5 | Initial review — 2 medium, 3 low findings, plus 3 additional body notes |
| 2 | 0 | New commit added missing `net/http` import in multi-instance E2E test (CI fix) — no source changes, no new findings |
| 3 | 0 | New commit disambiguated test credential fixtures and handled absent `parameters` key after strip — test-only, no source changes, no response to Round 1 findings |
| 4 | 0 | No new commits, no new findings |
| 5 | 0 | PR merged before any fixes to Round 1 feedback |

### Overall Assessment

The architecture of Phase 4 is well thought out: resolving the `connector_instance` selector once at request time and freezing `_connector_instance_id`/`_connector_instance_label` on the stored action JSON is the right approach — it decouples the approval record from mutable agent-connector state and gives every downstream execution path (standing auto-approve, pending approve, user-initiated SA execute, resource-detail resolver) a single authoritative source for routing. Test coverage is strong, especially the end-to-end fixture that verifies the correct credential actually reaches the connector on approve.

The open findings are all polish rather than correctness blockers: stricter input validation for the selector, logging the one Marshal failure mode, and a few cleanups. Worth addressing in a follow-up PR so the edges of the contract are as tight as the happy path.

---

PR #972, opened 2026-04-21
fix(connectors): default Google/Microsoft outbound email to HTML

### LINKED ISSUE #971: Emails sent via Google/Microsoft connectors render as plaintext instead of HTML
## Problem

When the agent sends emails through the Google (Gmail) and Microsoft (Graph) connectors, the body frequently arrives as plaintext — raw `<p>` / `<br>` tags show up in recipients' inboxes, and Markdown-ish bodies never render rich formatting.

### Gmail — always plaintext

`connectors/google/email_helpers.go:14` hard-codes the MIME header in `buildGmailRaw`:

```go
msg.WriteString("Content-Type: text/plain; charset=\"UTF-8\"\r\n")
```

Both `send_email` (`connectors/google/send_email.go:73`) and `send_email_reply` (`connectors/google/send_email_reply.go:108`) go through this helper, so **every** Gmail message is sent as `text/plain` regardless of body content. There's no HTML path at all.

### Microsoft Graph — weak auto-detection

`connectors/microsoft/send_email.go:87-90` calls `detectContentType(params.Body)`, which lives in `connectors/microsoft/validation.go:49-54`:

```go
func detectContentType(body string) string {
    if strings.Contains(body, "<") && strings.Contains(body, ">") {
        return "HTML"
    }
    return "Text"
}
```

This fires HTML only when both `<` and `>` appear anywhere in the body. Agent-written bodies without tags (Markdown, plain prose with `\n`) fall through to `Text`, which Graph renders without any formatting. It's also a surprising trigger when legitimate bodies happen to contain angle brackets (code snippets, comparisons).

## Proposed solution

Add an explicit `html: boolean` parameter to both connectors' email actions, **defaulting to `true`**, and make the schema description tell the agent that the body must be HTML when `html=true`.

### 1. Gmail — `connectors/google/send_email.go` + `send_email_reply.go` + `email_helpers.go`

- [ ] Add `HTML *bool` to `sendEmailParams` and `sendEmailReplyParams`. Treat nil as `true` (default HTML).
- [ ] When HTML, send `multipart/alternative` containing both a plain-text fallback (generated by stripping tags / decoding entities) and the HTML part. When `html=false`, keep the existing `text/plain` path.
- [ ] Update `buildGmailRaw` to accept the content mode and construct the multipart body with a generated boundary. Preserve the existing header sanitization (CRLF rejection for user input, `stripHeaderNewlines` for Gmail-sourced values).
- [ ] Unit tests covering: default (html), explicit `html=true` multipart structure, explicit `html=false` plaintext path, replies inherit the same behavior, and MIME header injection is still rejected.

### 2. Microsoft — `connectors/microsoft/send_email.go` + `validation.go`

- [ ] Add `HTML *bool` to `sendEmailParams`, defaulting to `true` when nil.
- [ ] Replace the `detectContentType` sniff: set Graph's `ContentType` to `"HTML"` or `"Text"` based on the param only. Remove the `<…>` heuristic entirely.
- [ ] Delete `detectContentType` if no other caller needs it. (`send_channel_message.go:72` and `create_calendar_event.go` both use it — migrate them to explicit params in the same PR, or leave a follow-up; see "Out of scope" below.)
- [ ] Unit tests: default (HTML), `html=false` → `Text`, spec JSON uses the param.

### 3. Connector manifests / OpenAPI spec

- [ ] Add the `html` parameter to the action definitions in the Google and Microsoft connector manifests. Include a description aimed at the agent, e.g.:
  > `html` (default `true`): when `true`, `body` must be valid HTML (use `<p>`, `<br>`, `<strong>`, etc. for formatting). When `false`, `body` is sent as plaintext and tags are not interpreted.
- [ ] Regenerate the OpenAPI spec / typed client so the agent surfaces the new param.

### 4. Post-task review

- [ ] Run `make test-backend` locally once (or note in PR that sandbox blocks it and rely on CI).
- [ ] `gofmt` all changed files.
- [ ] Verify no regression for existing plaintext callers by exercising `html=false` in tests.

## Out of scope (possible follow-ups)

- Migrating `microsoft.send_channel_message` and `microsoft.create_calendar_event` away from `detectContentType` to an explicit `html` param. They share the helper but have different UX considerations — file a separate issue if we don't tackle them here.
- Adding an `html` param to any other email-sending connector (e.g., SendGrid) that may have similar behavior.

## Acceptance criteria

- Sending an email via `google.send_email` / `google.send_email_reply` with default params renders as HTML in Gmail/Outlook/Apple Mail, with a usable plaintext fallback for clients that request it.
- Sending an email via `microsoft.send_email` with default params renders as HTML in Outlook.
- Passing `html: false` on either connector produces a pure plaintext message with tags not interpreted.
- The agent can discover the `html` parameter and its semantics from the connector schema / OpenAPI spec.

chiedocursor:
## Summary

Implements GitHub #971: outbound Gmail and Microsoft Graph emails now default to HTML rendering instead of plaintext or brittle tag sniffing.

- **Gmail (`google.send_email`, `google.send_email_reply`)**: Optional `html` (`*bool`, default when omitted: **true**). When HTML mode is on, messages are `multipart/alternative` with `text/html` plus a generated `text/plain` fallback (tags stripped, entities decoded). When `html: false`, behavior matches the previous single `text/plain` path.
- **Microsoft (`microsoft.send_email`)**: Same `html` flag. Graph `body.contentType` is **`HTML`** or **`Text`** from that flag only (default **HTML**).

**Parity (this update):** `microsoft.send_channel_message` and `microsoft.create_calendar_event` now use the same explicit `html` parameter and shared `graphBodyContentType` helper. The `<…>` tag heuristic (`detectContentType`) is removed.

Connector manifests and `connectors/microsoft/README.md` document the parameters.

## Tests

- `go test ./connectors/google/... ./connectors/microsoft/...`
- `go test ./connectors/...`

( Full `make test-backend` not run in this environment; CI should cover the rest. )

Closes #971

<div><a href="https://cursor.com/agents/bc-26ef83ec-76eb-4fe2-89d8-09a234c6b493"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-26ef83ec-76eb-4fe2-89d8-09a234c6b493"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/connectors/google/email_helpers.go
+++ b/connectors/google/email_helpers.go
@@ -1,17 +1,56 @@
 package google
 
 import (
+	"crypto/rand"
 	"encoding/base64"
+	"encoding/hex"
+	"html"
+	"regexp"
 	"strings"
 )
 
-// buildGmailRaw constructs a plain-text RFC 2822 email message and returns
-// it as a base64url-encoded string suitable for the Gmail API's raw field.
+var tagPattern = regexp.MustCompile(`<[^>]*>`)
+
+// emailHTMLDefault returns true when html is nil (default HTML) or *html is true.
+func emailHTMLDefault(html *bool) bool {
+	if html == nil {
+		return true
+	}
+	return *html
+}
+
+// htmlToPlainText produces a minimal plain-text fallback from HTML for the
+// text/plain MIME part (tags stripped, entities decoded).
+func htmlToPlainText(s string) string {
+	s = strings.ReplaceAll(s, "<br>", "\n")
+	s = strings.ReplaceAll(s, "<br/>", "\n")
+	s = strings.ReplaceAll(s, "<br />", "\n")
+	s = strings.ReplaceAll(s, "</p>", "\n")
+	s = strings.ReplaceAll(s, "</div>", "\n")
+	s = tagPattern.ReplaceAllString(s, "")
+	s = html.UnescapeString(s)
+	return strings.TrimSpace(s)
+}
+
+func randomMimeBoundary() string {
+	var b [12]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		// Extremely unlikely; use a fixed fallback (still unique enough for one message).
+		return "ps_mime_fallback_boundary"
+	}
+	return "ps_mime_" + hex.EncodeToString(b[:])
+}
+
+// buildGmailRaw constructs an RFC 2822 email message and returns it as a
+// base64url-encoded string suitable for the Gmail API's raw field.
+//
+// When html is true, the message is multipart/alternative with text/plain and
+// text/html parts. When html is false, a single text/plain part is used.
 //
 // Callers are responsible for sanitizing header values before passing them in.
 // extraHeaders are written between Subject and Content-Type, in the order given;
 // entries with an empty value are skipped.
-func buildGmailRaw(to, subject, body string, extraHeaders [][2]string) string {
+func buildGmailRaw(to, subject, body string, html bool, extraHeaders [][2]string) string {
 	var msg strings.Builder
 	msg.WriteString("To: " + to + "\r\n")
 	msg.WriteString("Subject: " + subject + "\r\n")
@@ -20,8 +59,30 @@ func buildGmailRaw(to, subject, body string, extraHeaders [][2]string) string {
 			msg.WriteString(h[0] + ": " + h[1] + "\r\n")
 		}
 	}
-	msg.WriteString("Content-Type: text/plain; charset=\"UTF-8\"\r\n")
-	msg.WriteString("\r\n")
-	msg.WriteString(body)
+	if html {
+		boundary := randomMimeBoundary()
+		plain := htmlToPlainText(body)
+		if plain == "" {
+			plain = " "
+		}
+		msg.WriteString("MIME-Version: 1.0\r\n")
+		msg.WriteString("Content-Type: multipart/alternative; boundary=\"" + boundary + "\"\r\n")
+		msg.WriteString("\r\n")
+		msg.WriteString("--" + boundary + "\r\n")
+		msg.WriteString("Content-Type: text/plain; charset=\"UTF-8\"\r\n")
+		msg.WriteString("\r\n")
+		msg.WriteString(plain)
+		msg.WriteString("\r\n")
+		msg.WriteString("--" + boundary + "\r\n")
+		msg.WriteString("Content-Type: text/html; charset=\"UTF-8\"\r\n")
+		msg.WriteString("\r\n")
+		msg.WriteString(body)
+		msg.WriteString("\r\n")
+		msg.WriteString("--" + boundary + "--\r\n")
+	} else {
+		msg.WriteString("Content-Type: text/plain; charset=\"UTF-8\"\r\n")
+		msg.WriteString("\r\n")
+		msg.WriteString(body)
+	}
 	return base64.RawURLEncoding.EncodeToString([]byte(msg.String()))
 }
--- a/connectors/microsoft/manifest_templates.go
+++ b/connectors/microsoft/manifest_templates.go
@@ -49,7 +49,7 @@ func microsoftTemplates() []connectors.ManifestTemplate {
 			ActionType:  "microsoft.create_calendar_event",
 			Name:        "Create calendar events",
 			Description: "Agent can create events on the calendar with any details.",
-			Parameters:  json.RawMessage(`{"subject":"*","start":"*","end":"*","time_zone":"*","body":"*","attendees":"*","location":"*","calendar_id":""}`),
+			Parameters:  json.RawMessage(`{"subject":"*","start":"*","end":"*","time_zone":"*","body":"*","attendees":"*","location":"*","calendar_id":"","html":"*"}`),
 		},
 		// --- OneDrive read ---
 		{
@@ -209,7 +209,7 @@ func microsoftTemplates() []connectors.ManifestTemplate {
 			ActionType:  "microsoft.send_channel_message",
 			Name:        "Send channel messages",
 			Description: "Agent can send messages to any Teams channel.",
-			Parameters:  json.RawMessage(`{"team_id":"*","channel_id":"*","message":"*","reply_to_message_id":"*"}`),
+			Parameters:  json.RawMessage(`{"team_id":"*","channel_id":"*","message":"*","reply_to_message_id":"*","html":"*"}`),
 		},
 	}
 }
--- a/connectors/microsoft/microsoft.go
+++ b/connectors/microsoft/microsoft.go
@@ -160,7 +160,12 @@ func (c *MicrosoftConnector) Manifest() *connectors.ConnectorManifest {
 						},
 						"body": {
 							"type": "string",
-							"description": "Event body/description (HTML supported)"
+							"description": "Event body/description. When html is true (default), use valid HTML. When false, plain text only."
+						},
+						"html": {
+							"type": "boolean",
+							"default": true,
+							"description": "When true (default), body is HTML. When false, body is plain text only."
 						},
 						"attendees": {
 							"type": "array",
@@ -441,7 +446,12 @@ func (c *MicrosoftConnector) Manifest() *connectors.ConnectorManifest {
 						},
 						"message": {
 							"type": "string",
-							"description": "Message content (HTML or plain text — auto-detected)"
+							"description": "Message content. When html is true (default), use valid HTML. When false, plain text only."
+						},
+						"html": {
+							"type": "boolean",
+							"default": true,
+							"description": "When true (default), message is HTML. When false, plain text only."
 						},
 						"reply_to_message_id": {
 							"type": "string",
--- a/connectors/microsoft/microsoft.go
+++ b/connectors/microsoft/microsoft.go
@@ -4,9 +4,9 @@
 package microsoft
 
 import (
-	_ "embed"
 	"bytes"
 	"context"
+	_ "embed"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -62,6 +62,7 @@ func (c *MicrosoftConnector) ID() string { return "microsoft" }
 
 // Manifest returns the connector's metadata manifest. Used by the server to
 // auto-seed DB rows on startup, replacing manual seed.go files.
+//
 //go:embed logo.svg
 var logoSVG string
 
@@ -93,12 +94,17 @@ func (c *MicrosoftConnector) Manifest() *connectors.ConnectorManifest {
 						},
 						"body": {
 							"type": "string",
-							"description": "Email body (HTML or plain text — auto-detected)"
+							"description": "Email body. When html is true (default), use valid HTML (e.g. <p>, <br>, <strong>). When html is false, body is sent as plain text and tags are not interpreted."
 						},
 						"cc": {
 							"type": "array",
 							"items": {"type": "string"},
 							"description": "CC recipient email addresses"
+						},
+						"html": {
+							"type": "boolean",
+							"default": true,
+							"description": "When true (default), body must be valid HTML. When false, body is sent as plain text only."
 						}
 					}
 				}`)),
@@ -213,7 +219,7 @@ func (c *MicrosoftConnector) Manifest() *connectors.ConnectorManifest {
 				}`)),
 			},
 			{
-			ActionType:  "microsoft.list_drive_files",
+				ActionType:  "microsoft.list_drive_files",
 				Name:        "List Drive Files",
 				Description: "List files and folders in OneDrive",
 				RiskLevel:   "low",
--- a/connectors/microsoft/validation.go
+++ b/connectors/microsoft/validation.go
@@ -53,6 +53,15 @@ func detectContentType(body string) string {
 	return "Text"
 }
 
+// sendMailGraphContentType maps the explicit html flag to Microsoft Graph
+// sendMail body contentType. Nil defaults to HTML (issue #971).
+func sendMailGraphContentType(html *bool) string {
+	if html != nil && !*html {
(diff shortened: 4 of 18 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---
*Watch session ended after 10 minutes of inactivity. Processed 0 comments across 10 poll cycles.*

## 🔧 CI / audit remediation

### Remediation round 1

- **ci.yml** — conclusion `failure` — Backend Tests failed: TestGetProfile_TraceIDPresent expected trace_id in JSON but NewRouter tests omit TraceIDMiddleware (only main.go wraps the mux). Wrapped the test handler with TraceIDMiddleware to match production. (`fbaec620`)

connectors/microsoft/manifest_templates.go
```diff
@@ -49,7 +49,7 @@ func microsoftTemplates() []connectors.ManifestTemplate {
 			ActionType:  "microsoft.create_calendar_event",
 			Name:        "Create calendar events",
 			Description: "Agent can create events on the calendar with any details.",
-			Parameters:  json.RawMessage(`{"subject":"*","start":"*","end":"*","time_zone":"*","body":"*","attendees":"*","location":"*","calendar_id":""}`),
+			Parameters:  json.RawMessage(`{"subject":"*","start":"*","end":"*","time_zone":"*","body":"*","attendees":"*","location":"*","calendar_id":"","html":"*"}`),
```
chiedoclaude, line 52: **[Critical · Correctness]** The PR added `"html":"*"` here and to `tpl_microsoft_send_channel_message`, but missed every other template that permits an action whose schema now accepts `html`. Un-updated templates in this file:

- `tpl_microsoft_send_email` — `{"to":"*","subject":"*","body":"*"}`
- `tpl_microsoft_send_email_to_domain` — `{"to":{"$pattern":"*@example.com"},"subject":"*","body":"*"}`

And in `connectors/google/manifest_templates.go`:

- `tpl_google_send_email`
- `tpl_google_send_email_to_recipient`
- `tpl_google_send_email_to_domain`
- `tpl_google_send_email_reply`

**Why it matters:** `db/action_config_validate.go:247-254` (`ValidateParametersAgainstConfig`) rejects any execution parameter that is not listed in the template's `Parameters` with `"parameter not defined in configuration"`. Since HTML is now the default, the *only* way to send plaintext is `html: false` — and for agents bound to these templates that call will be **blocked by permission validation before the request ever reaches the connector**. Net effect: customers who were relying on plaintext (including bodies that contain `<` or `&`, which used to be sent as-is) now have:

1. No opt-out for the new HTML default, and
2. Their existing plaintext content silently rendered as HTML, with bare `<`/`&` producing broken output in recipients' inboxes.

**Fix:** Add `"html":"*"` to all six templates above (or drop the default change to HTML until the templates are in sync). E.g.:

```go
// tpl_microsoft_send_email
Parameters: json.RawMessage(`{"to":"*","subject":"*","body":"*","cc":"*","html":"*"}`),
```

A regression test that instantiates each of these templates and calls `ValidateParametersAgainstConfig` with `{"html":false}` would prevent this from happening again.

---

connectors/microsoft/validation.go
```diff
@@ -45,12 +45,14 @@ func validateItemID(id string) error {
...
 
-// detectContentType returns "HTML" if the body contains HTML tags, otherwise "Text".
-func detectContentType(body string) string {
-	if strings.Contains(body, "<") && strings.Contains(body, ">") {
-		return "HTML"
+// graphBodyContentType maps the explicit html flag to Microsoft Graph
+// body contentType (sendMail, Teams messages, event body, etc.). Nil defaults
+// to HTML so Markdown-ish content without angle brackets still renders as HTML.
+func graphBodyContentType(html *bool) string {
+	if html != nil && !*html {
```
chiedoclaude, line 52: **[High · Correctness]** This function always returns `"HTML"` when `html` is nil, which is a hard behavior change for existing callers. The previous `detectContentType` returned `"Text"` for bodies with no `<>` characters; now even plain prose like `"Report is 3 < 5 items"` or `"Fees & charges"` is marked `HTML` and passed through to Graph verbatim.

**Why it matters:** Microsoft Graph interprets that content as HTML in the recipient's client:
- `"3 < 5 items"` — the browser sees `< 5 items"` as an unterminated tag and swallows everything to the next `>`.
- `"Tom & Jerry"` — `&` followed by a word boundary will be treated as a malformed entity by some clients.

Existing agents that were happily sending plaintext have no awareness the default flipped, and they'll only find out via rendering bugs in production.

**Fix options (pick one):**
1. Keep the default as `Text` unless the caller opts into HTML (invert the flag name to `plain *bool` with default true, or rename to `content_type: "html" | "text"` and default text — safest migration).
2. Default to HTML but auto-escape the body when `html` is omitted and the body contains no HTML tags, so legacy plaintext-style content renders identically to before.
3. Accept the breaking change but update the PR title/body from `fix(connectors)` to `feat!(connectors):` and add a note in the changelog / release notes so operators know to audit agent prompts.

Either way, please add an explicit callout in the PR description (and `connectors/microsoft/README.md`) that this is a breaking default change for existing send_email, create_calendar_event, and send_channel_message callers.

---

connectors/google/email_helpers.go
```diff
@@ -1,17 +1,56 @@
...
+	s = tagPattern.ReplaceAllString(s, "")
+	s = html.UnescapeString(s)
+	return strings.TrimSpace(s)
+}
+
+func randomMimeBoundary() string {
+	var b [12]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		// Extremely unlikely; use a fixed fallback (still unique enough for one message).
+		return "ps_mime_fallback_boundary"
```
chiedoclaude, line 39: **[High · Security]** If `rand.Read` ever returns an error, every message that hits this branch uses the same constant boundary `"ps_mime_fallback_boundary"`. The comment "(still unique enough for one message)" is factually wrong — it's identical across every message.

**Why it matters:** In HTML mode the user-controlled `body` is written verbatim between MIME boundary markers. If an attacker controls body content and knows the boundary, they can inject a counterfeit MIME part — e.g. a second `text/html` alternative with attacker-chosen content, or an unrelated attachment part. With a random 96-bit boundary that's infeasible; with a hard-coded fallback it's trivial. `rand.Read` from `crypto/rand` effectively never fails, but "effectively never" isn't a safe guarantee to rely on in a mail-sending code path: fd exhaustion, seccomp filters, and sandboxing can all cause it.

**Fix:** Don't silently degrade. Either bubble the error up (change the signature to return `(string, error)`), or `panic` — this is an unrecoverable system failure, not a routine condition:

```go
func randomMimeBoundary() string {
    var b [12]byte
    if _, err := rand.Read(b[:]); err != nil {
        panic(fmt.Errorf("crypto/rand.Read failed: %w", err))
    }
    return "ps_mime_" + hex.EncodeToString(b[:])
}
```

If you prefer to keep the string return, propagate the error to `buildGmailRaw` and then to the action's `Execute`, which can surface a proper `ExternalError`.

---

connectors/google/email_helpers.go
```diff
@@ -1,17 +1,56 @@
...
+func emailHTMLDefault(html *bool) bool {
+	if html == nil {
+		return true
+	}
+	return *html
+}
+
+// htmlToPlainText produces a minimal plain-text fallback from HTML for the
+// text/plain MIME part (tags stripped, entities decoded).
+func htmlToPlainText(s string) string {
```
chiedoclaude, line 24: **[Medium · Correctness]** This plain-text fallback is quite lossy on realistic HTML inputs — the regex strips tags but doesn't strip their *content* for script/style blocks, loses list structure, and drops link targets:

- `<script>alert(1)</script>` → `alert(1)` in the text/plain part (script body leaks).
- `<style>p{color:red}</style>Hi` → `p{color:red}Hi`.
- `<ul><li>A</li><li>B</li></ul>` → `AB` (no separators).
- `<a href="https://x.com">click</a>` → `click` (URL lost — bad for accessibility and spam-scoring).
- Uppercase `<BR>`, `<P>`, `<DIV>` aren't matched by the literal `ReplaceAll` calls, so they get stripped without inserting newlines.

**Why it matters:** The text/plain part is what users see when their client doesn't render HTML (some corporate filters, plain-text preferences, screen readers). A malformed or unreadable fallback hurts deliverability (spam filters penalize mismatched plain/HTML alternatives) and accessibility. It's also what some anti-phishing systems scan.

**Fix:** Either (a) document the limitation and state that callers should provide their own plaintext alternative when quality matters, or (b) use `golang.org/x/net/html` to walk the DOM and emit a proper conversion. At minimum:

```go
// Strip <script> and <style> with their content first.
var scriptStyle = regexp.MustCompile(`(?is)<(script|style)[^>]*>.*?</\1>`)

func htmlToPlainText(s string) string {
    s = scriptStyle.ReplaceAllString(s, "")
    // case-insensitive block-level replacements
    s = regexp.MustCompile(`(?i)<br\s*/?>`).ReplaceAllString(s, "\n")
    s = regexp.MustCompile(`(?i)</(p|div|li)>`).ReplaceAllString(s, "\n")
    s = regexp.MustCompile(`(?i)<li[^>]*>`).ReplaceAllString(s, "- ")
    s = tagPattern.ReplaceAllString(s, "")
    return strings.TrimSpace(html.UnescapeString(s))
}
```

---

connectors/google/email_helpers.go
```diff
@@ -1,17 +1,56 @@
...
+	"regexp"
 	"strings"
 )
 
-// buildGmailRaw constructs a plain-text RFC 2822 email message and returns
-// it as a base64url-encoded string suitable for the Gmail API's raw field.
+var tagPattern = regexp.MustCompile(`<[^>]*>`)
+
+// emailHTMLDefault returns true when html is nil (default HTML) or *html is true.
+func emailHTMLDefault(html *bool) bool {
```
chiedoclaude, line 15: **[Medium · Maintainability]** The parameter name `html` shadows the `html` standard library package imported on line 7, which is then used as `html.UnescapeString(s)` a few lines below inside `htmlToPlainText`. Readers scanning the file see two meanings of `html` in the same ~30 lines. The same shadow repeats in `buildGmailRaw(to, subject, body string, html bool, ...)`.

**Why it matters:** It compiles fine today (the shadow is scoped to each function), but the next person to add `html.EscapeString` or similar inside `emailHTMLDefault`/`buildGmailRaw` will get a confusing "html.EscapeString undefined (type *bool has no field or method EscapeString)" error and have to untangle the shadow. It also makes the code harder to grep for stdlib usage.

**Fix:** Rename the parameter — `isHTML` or `htmlMode` reads naturally and keeps the package accessible:

```go
func emailHTMLDefault(isHTML *bool) bool {
    if isHTML == nil {
        return true
    }
    return *isHTML
}

func buildGmailRaw(to, subject, body string, isHTML bool, extraHeaders [][2]string) string {
```

Same rename would be nice in `connectors/microsoft/validation.go` `graphBodyContentType(html *bool)`, though that file doesn't import the `html` package so it's purely a consistency point.

---

connectors/microsoft/microsoft.go
```diff
@@ -4,9 +4,9 @@
 package microsoft
 
 import (
-	_ "embed"
 	"bytes"
```
chiedoclaude, line 7: **[Low · Maintainability]** These two unrelated changes — moving `_ "embed"` into alphabetical order and fixing the indentation of `ActionType: "microsoft.list_drive_files"` at line ~227 — are orthogonal to the email/HTML work in this PR. Per `CLAUDE.md` ("Avoid reformatting or reordering existing code"), these belong in a separate cleanup commit.

**Why it matters:** Touching unrelated lines inflates the diff, creates merge-conflict surface with other in-flight branches, and makes `git blame` noisier when someone later tries to understand why the embed import moved.

**Fix:** Revert these two hunks. They're correct improvements — just not on this PR.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 6 (1 critical, 2 high, 2 medium, 1 low)

The core intent — replacing the brittle `<…>` tag-sniffing heuristic with an explicit `html` flag that defaults to HTML — is a good call, and the test coverage across both connectors is solid. However, there's a **critical gap in the permission-slip template coverage**: the new `html` parameter was added to manifest schemas but only to 2 of the ~8 affected templates in `connectors/microsoft/manifest_templates.go` and `connectors/google/manifest_templates.go`. Because `db/action_config_validate.go:247-254` strictly rejects any parameter not listed in the template's `Parameters` JSON, **agents using the un-updated templates have no way to opt out of the new HTML default** — passing `html: false` will fail validation, and bodies containing `<` or `&` will mis-render in recipients' clients. This is a silent regression for every plaintext-using customer.

Additionally, a few smaller issues around the deterministic MIME-boundary fallback, the naive `htmlToPlainText` helper, and some unrelated whitespace churn.

### What Looks Good
- Replacing `detectContentType` with an explicit flag is the right long-term move.
- Test coverage is thorough: explicit-true, explicit-false, default, and multipart structure are all exercised across Gmail, Graph sendMail, Teams channel messages, and calendar events.
- Header-injection safeguards (`containsNewline`, `stripHeaderNewlines`) are preserved.
- Random MIME boundary using `crypto/rand` is the right primitive.
- README parameter tables were updated to match the new schema.

---

---

Issue #974, opened 2026-04-21
Remove credential instance labels; replace UI with per-credential checkboxes


chiedoclaude:
## Background

We recently added support for multiple credential instances per connector. Each instance requires a user-supplied `label` (e.g., "Supersuit", "Innovation Hub"). In practice this is redundant — OAuth connections already carry a workspace name and API-key credentials already carry a user-given nickname — so users are being asked to name the same thing twice.

This issue removes the label concept and reshapes the per-agent connector UI around a simpler mental model: **"which of the available credentials is this agent allowed to use?"** — rendered as a list of checkboxes, with one credential marked default.

## Decisions (already resolved)

- **Default selection:** keep the existing `is_default` flag and default-per-pair constraint. The checkbox list shows one credential as default (star/radio).
- **Display name source:** use the credential's own nickname (API keys already have a user-given name; OAuth connections carry workspace/team name).
- **Legacy data:** drop the `label` column outright. No backfill of historical `_connector_instance_label` on old action rows — old audit records retain whatever was frozen into their action JSON.
- **Manage credentials** experience stays as-is; this issue only changes the per-agent assignment UI.

## Phase 1 — Database

- [ ] Create migration to drop `agent_connectors.label`
- [ ] Drop unique index `uq_agent_connectors_agent_connector_label`
- [ ] Drop the trigger that auto-populates `label` from `connectors.name`
- [ ] Keep `connector_instance_id`, `is_default`, and `uq_agent_connectors_default_per_pair` — all still needed
- [ ] Update the seed file to reflect the new schema

## Phase 2 — Backend (can run in parallel with Phase 3 once Phase 1 is merged)

- [ ] Remove `Label` field, `MaxAgentConnectorInstanceLabelLen`, and `AgentConnectorInstanceErrDuplicateLabel` from `db/agent_connector_instances.go`
- [ ] `POST /agents/{agent_id}/connectors/{connector_id}/instances`: drop `label` from request body; creation now just allocates an instance row and (optionally) assigns a credential
- [ ] `PATCH /agents/{agent_id}/connectors/{connector_id}/instances/{instance_id}`: remove label rename path; keep `is_default` toggle
- [ ] `ResolveAgentConnectorInstance()` (`db/agent_connector_instances.go`): accept UUID or **credential nickname** instead of label for the `connector_instance` approval parameter
[...]

PR #979, opened 2026-04-21
Phase 1: drop agent_connectors.label (migration + DB/API alignment)

chiedocursor:
## Summary

Implements [issue #974](https://github.com/supersuit-tech/permission-slip/issues/974) **Phases 1–2**:

### Phase 1 (database)
- Migration `20260421134629_drop_agent_connectors_label.sql`: drop `label`, unique index on label, replace insert trigger (first-row default only).
- `EnableAgentConnector` idempotent without label unique constraint.
- Instance display derived from credential binding (API + DB).

### Phase 2 (backend + spec)
- `AgentConnectorInstance` / JSON: `display` (from credential nickname / OAuth name); struct field `DisplayName` in Go.
- Capabilities `connectors[].instances[]`: `display` instead of `label`.
- Approvals: freeze `_connector_instance_display` on action JSON (legacy `_connector_instance_label` still in audit redaction for old rows).
- `available_instances` error detail uses credential display strings.
- OpenAPI updated + `openapi.bundle.yaml` regenerated (`make bundle`).

### Clients (required for spec + compile)
- **Frontend:** connector instances UI — add instance without label, remove rename; approval dialog reads `_connector_instance_display` with fallback to `_connector_instance_label`.
- **Mobile:** `connectorInstanceLabelFromAction` prefers `_connector_instance_display`.

Ref #974

## Test plan

- [ ] CI: `make test-backend`, `make test-frontend`, `make mobile-test`
- Local agent: frontend vitest (connector hooks + section + approvalConnectorLabel); mobile jest (approvalUtils) — passed
- Backend `go test` not run locally (toolchain constraint in agent env)

## Risks / notes

- PATCH connector instance only supports `is_default` (no rename).
- Empty credential binding → empty `display` (omitempty in JSON).

<div><a href="https://cursor.com/agents/bc-7f53cbe6-8fa6-4f64-b0c4-e9791215155a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7f53cbe6-8fa6-4f64-b0c4-e9791215155a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/db/agent_connector_instances.go
+++ b/db/agent_connector_instances.go
@@ -4,21 +4,13 @@ import (
 	"context"
 	"errors"
 	"strings"
-	"time"
-	"unicode/utf8"
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
-	"github.com/jackc/pgx/v5/pgconn"
 )
 
-// ErrAgentConnectorInstanceLabelRequired is returned when label is empty after trim.
-var ErrAgentConnectorInstanceLabelRequired = errors.New("label is required")
-
-// ErrAgentConnectorInstanceLabelTooLong is returned when label exceeds MaxAgentConnectorInstanceLabelLen runes.
-var ErrAgentConnectorInstanceLabelTooLong = errors.New("label exceeds maximum length")
-
 // AgentConnectorInstance is a row in agent_connectors (one instance of a connector type for an agent).
+// Label is a display string derived from the bound credential (API key label or OAuth workspace name).
 type AgentConnectorInstance struct {
 	ConnectorInstanceID string
 	AgentID             int64
@@ -41,33 +33,35 @@ type AgentConnectorInstanceErrCode string
 
 const (
 	AgentConnectorInstanceErrNotFound            AgentConnectorInstanceErrCode = "connector_instance_not_found"
-	AgentConnectorInstanceErrDuplicateLabel      AgentConnectorInstanceErrCode = "duplicate_instance_label"
 	AgentConnectorInstanceErrCannotDeleteDefault AgentConnectorInstanceErrCode = "cannot_delete_default_instance"
 )
 
-// MaxAgentConnectorInstanceLabelLen is the maximum rune length for instance labels (API + DB).
-const MaxAgentConnectorInstanceLabelLen = 256
+// agentConnectorInstanceSelect is the standard SELECT for an agent_connectors row with display label
+// from the assigned credential (static API key label or OAuth extra_data name).
+const agentConnectorInstanceSelect = `
+	SELECT ac.connector_instance_id, ac.agent_id, ac.connector_id, ac.approver_id,
+	       COALESCE(cr.label, oc.extra_data->>'name', '') AS label,
+	       ac.is_default, ac.enabled_at
+	FROM agent_connectors ac
+	LEFT JOIN agent_connector_credentials acc
+	       ON acc.agent_id = ac.agent_id
+	      AND acc.connector_id = ac.connector_id
+	      AND acc.approver_id = ac.approver_id
+	      AND acc.connector_instance_id = ac.connector_instance_id
+	LEFT JOIN credentials cr ON cr.id = acc.credential_id
+	LEFT JOIN oauth_connections oc ON oc.id = acc.oauth_connection_id`
 
 // CreateAgentConnectorInstanceParams holds parameters for creating a new instance (non-default row).
 type CreateAgentConnectorInstanceParams struct {
 	AgentID     int64
 	ApproverID  string
 	ConnectorID string
-	Label       string
 }
 
-// CreateAgentConnectorInstance inserts a new agent connector instance with the given label.
+// CreateAgentConnectorInstance inserts a new agent connector instance.
 // The first instance for an (agent, connector) pair is the default (enforced by DB trigger);
 // additional instances are non-default.
 func CreateAgentConnectorInstance(ctx context.Context, db DBTX, p CreateAgentConnectorInstanceParams) (*AgentConnectorInstance, error) {
-	label := strings.TrimSpace(p.Label)
-	if label == "" {
-		return nil, ErrAgentConnectorInstanceLabelRequired
-	}
-	if utf8.RuneCountInString(label) > MaxAgentConnectorInstanceLabelLen {
-		return nil, ErrAgentConnectorInstanceLabelTooLong
-	}
-
 	var agentOK bool
 	if err := db.QueryRow(ctx,
 		`SELECT EXISTS (SELECT 1 FROM agents WHERE agent_id = $1 AND approver_id = $2)`,
@@ -104,21 +98,17 @@ func CreateAgentConnectorInstance(ctx context.Context, db DBTX, p CreateAgentCon
 		return nil, &AgentConnectorError{Code: AgentConnectorErrConnectorNotEnabled}
 	}
 
-	row := db.QueryRow(ctx, `
-		INSERT INTO agent_connectors (agent_id, approver_id, connector_id, label, is_default)
-		VALUES ($1, $2, $3, $4, false)
-		RETURNING connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at`,
-		p.AgentID, p.ApproverID, p.ConnectorID, label,
-	)
-	inst, err := scanAgentConnectorInstance(row)
+	var newID string
+	err := db.QueryRow(ctx, `
+		INSERT INTO agent_connectors (agent_id, approver_id, connector_id, is_default)
+		VALUES ($1, $2, $3, false)
+		RETURNING connector_instance_id::text`,
+		p.AgentID, p.ApproverID, p.ConnectorID,
+	).Scan(&newID)
 	if err != nil {
-		var pgErr *pgconn.PgError
-		if errors.As(err, &pgErr) && pgErr.Code == PgCodeUniqueViolation {
-			return nil, &AgentConnectorInstanceError{Code: AgentConnectorInstanceErrDuplicateLabel}
-		}
 		return nil, err
 	}
-	return inst, nil
+	return GetAgentConnectorInstance(ctx, db, p.AgentID, p.ApproverID, p.ConnectorID, newID)
 }
 
 func scanAgentConnectorInstance(row pgx.Row) (*AgentConnectorInstance, error) {
@@ -135,11 +125,9 @@ func scanAgentConnectorInstance(row pgx.Row) (*AgentConnectorInstance, error) {
 
 // ListAgentConnectorInstances returns all instances for an agent+connector scoped to the approver.
 func ListAgentConnectorInstances(ctx context.Context, db DBTX, agentID int64, approverID, connectorID string) ([]AgentConnectorInstance, error) {
-	rows, err := db.Query(ctx, `
-		SELECT connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at
-		FROM agent_connectors
-		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3
-		ORDER BY enabled_at ASC, connector_instance_id ASC`,
+	rows, err := db.Query(ctx, agentConnectorInstanceSelect+`
+		WHERE ac.agent_id = $1 AND ac.approver_id = $2 AND ac.connector_id = $3
+		ORDER BY ac.enabled_at ASC, ac.connector_instance_id ASC`,
 		agentID, approverID, connectorID,
 	)
 	if err != nil {
@@ -149,24 +137,19 @@ func ListAgentConnectorInstances(ctx context.Context, db DBTX, agentID int64, ap
 
 	var out []AgentConnectorInstance
 	for rows.Next() {
-		var inst AgentConnectorInstance
-		if err := rows.Scan(
-			&inst.ConnectorInstanceID, &inst.AgentID, &inst.ConnectorID, &inst.ApproverID,
-			&inst.Label, &inst.IsDefault, &inst.EnabledAt,
-		); err != nil {
+		inst, err := scanAgentConnectorInstance(rows)
+		if err != nil {
 			return nil, err
 		}
-		out = append(out, inst)
+		out = append(out, *inst)
 	}
 	return out, rows.Err()
 }
 
 // GetAgentConnectorInstance returns a single instance by ID, scoped to agent and approver.
 func GetAgentConnectorInstance(ctx context.Context, db DBTX, agentID int64, approverID, connectorID, connectorInstanceID string) (*AgentConnectorInstance, error) {
-	row := db.QueryRow(ctx, `
-		SELECT connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at
-		FROM agent_connectors
-		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3 AND connector_instance_id = $4::uuid`,
+	row := db.QueryRow(ctx, agentConnectorInstanceSelect+`
+		WHERE ac.agent_id = $1 AND ac.approver_id = $2 AND ac.connector_id = $3 AND ac.connector_instance_id = $4::uuid`,
 		agentID, approverID, connectorID, connectorInstanceID,
 	)
 	inst, err := scanAgentConnectorInstance(row)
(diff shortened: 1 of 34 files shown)


db/agent_connector_instances.go
```diff
@@ -4,27 +4,19 @@ import (
...
-
 // AgentConnectorInstance is a row in agent_connectors (one instance of a connector type for an agent).
+// DisplayName is derived from the bound credential (API key label or OAuth workspace name in extra_data).
 type AgentConnectorInstance struct {
 	ConnectorInstanceID string
 	AgentID             int64
 	ConnectorID         string
 	ApproverID          string
-	Label               string
+	DisplayName         string
```
chiedoclaude, line 20: **[Critical · Correctness]** The `"time"` import was removed from this file (see top of diff), but `EnabledAt time.Time` on line 21 still references it. This file will fail to compile (`undefined: time`), and because every other package in this PR imports `db`, the whole backend build will break.

**Why:** The PR notes call out that `go build`/`go test` can't run in the agent sandbox, so this slipped through. CI will catch it, but it means the claim "Phase 1 complete" is premature — nothing after this commit actually builds.

**Fix:** Restore the import.

```go
import (
	"context"
	"errors"
	"strings"
	"time"

	"github.com/google/uuid"
	"github.com/jackc/pgx/v5"
)
```

---

db/agent_connectors.go
```diff
@@ -100,14 +100,17 @@ func EnableAgentConnector(ctx context.Context, db DBTX, agentID int64, approverI
 			INSERT INTO agent_connectors (agent_id, approver_id, connector_id)
 			SELECT $1, $2, $3
 			WHERE EXISTS (SELECT 1 FROM agent_check)
-			ON CONFLICT (agent_id, connector_id, label) DO NOTHING
+			  AND NOT EXISTS (
+			      SELECT 1 FROM agent_connectors ac0
+			      WHERE ac0.agent_id = $1 AND ac0.approver_id = $2 AND ac0.connector_id = $3
+			  )
```
chiedoclaude, line 122: **[High · Correctness]** Replacing `ON CONFLICT (agent_id, connector_id, label) DO NOTHING` with a `NOT EXISTS` sub-select breaks idempotency under concurrency.

After dropping the `label` column and the `uq_agent_connectors_agent_connector_label` unique index, nothing prevents two concurrent `EnableAgentConnector` callers from both passing the `NOT EXISTS` check, each minting its own `connector_instance_id` (from the column default), and both attempting the INSERT:

- Both hit the BEFORE INSERT trigger, which sets `is_default = true` for both (neither sees the other's uncommitted row).
- One INSERT succeeds; the other fails on `uq_agent_connectors_default_per_pair`. The losing caller returns a raw Postgres unique-violation error — not the idempotent success the API contract promises.
- On some isolation paths you can also end up with two rows for the same `(agent_id, approver_id, connector_id)` where only one is default, which is an invariant violation the old unique index explicitly prevented.

**Why it matters:** `EnableAgentConnector` is called on every agent start-up / capability registration path; a retry storm or a user double-clicking “Enable” can hit this. The symptom will be an intermittent 500 and — worse — silently split state.

**Fix:** use a predicate-unique index as the ON CONFLICT target so the INSERT stays atomic:

```sql
CREATE UNIQUE INDEX uq_agent_connectors_default_per_pair
    ON agent_connectors (agent_id, connector_id) WHERE is_default;
```

already exists. You can write the idempotent insert against it:

```sql
INSERT INTO agent_connectors (agent_id, approver_id, connector_id, is_default)
VALUES ($1, $2, $3, true)
ON CONFLICT (agent_id, connector_id) WHERE is_default DO NOTHING
RETURNING agent_id, connector_id, enabled_at
```

(and drop the trigger's auto-default path for this call site, or keep the trigger and use `ON CONFLICT DO NOTHING` without a target — PG will pick the violated constraint).

---

db/agent_connector_instances.go
```diff
@@ -361,10 +310,9 @@ func ResolveAgentConnectorInstance(ctx context.Context, db DBTX, agentID int64,
 		return GetAgentConnectorInstance(ctx, db, agentID, approverID, connectorID, sel)
 	}
 
-	row := db.QueryRow(ctx, `
-		SELECT connector_instance_id, agent_id, connector_id, approver_id, label, is_default, enabled_at
-		FROM agent_connectors
-		WHERE agent_id = $1 AND approver_id = $2 AND connector_id = $3 AND label = $4`,
+	row := db.QueryRow(ctx, agentConnectorInstanceSelect+`
+		WHERE ac.agent_id = $1 AND ac.approver_id = $2 AND ac.connector_id = $3
+		  AND COALESCE(cr.label, oc.extra_data->>'name', '') = $4`,
```
chiedoclaude: **[High · Correctness]** `ResolveAgentConnectorInstance` can now silently route to the wrong instance.

Before: `agent_connectors (agent_id, connector_id, label)` was unique, so selector → instance was 1:1.

After: the match is `COALESCE(cr.label, oc.extra_data->>'name', '') = $4` across a LEFT JOIN to credentials / oauth_connections. Nothing guarantees uniqueness of that computed string within `(agent, connector)`:

- Two OAuth connections can have the same `extra_data->>'name'` (e.g., two Slack workspaces that happen to share a display name, or the same workspace re-connected on two instances).
- Nothing stops a user from binding the same credential (or two same-labeled credentials) to two different instances.
- All instances without a bound credential collapse to `''`.

`db.QueryRow` then returns the first matching row and ignores the rest — so a request for `connector_instance: "Engineering"` can execute against whichever row Postgres happens to sort first. For an approval system whose job is *correctness* of action routing, this is the worst kind of silent failure.

**Fix:** switch to `db.Query`, and if the result set has >1 rows for a non-UUID selector, return a well-typed ambiguous-selector error (HTTP 400) listing the matching instance UUIDs. The API already has the machinery for "multiple instances, pick one" (see `applyConnectorInstanceToAction`) — reuse it. Also add a unit test for two instances with the same display.

---

db/migrations/20260421134629_drop_agent_connectors_label.sql
```diff
@@ -0,0 +1,80 @@
...
+UPDATE agent_connectors ac
+SET label = c.name
+FROM connectors c
+WHERE c.id = ac.connector_id;
+
+ALTER TABLE agent_connectors
+    ALTER COLUMN label SET NOT NULL;
+
+CREATE UNIQUE INDEX uq_agent_connectors_agent_connector_label
+    ON agent_connectors (agent_id, connector_id, label);
```
chiedoclaude, line 51: **[High · Migration Safety]** The Down migration will fail on any database that has multiple instances per `(agent, connector)`.

The UPDATE sets every row's `label` to `connectors.name`, so for an agent with two Slack instances both rows get `label = 'Slack'`. The subsequent `CREATE UNIQUE INDEX uq_agent_connectors_agent_connector_label ON agent_connectors (agent_id, connector_id, label)` then errors with a duplicate-key violation and the rollback aborts mid-transaction.

**Why it matters:** A broken Down migration is effectively no Down migration — if Phase 1 needs to be rolled back in production (and it might, given the other correctness issues in this PR), goose can't do it cleanly. Operators end up hand-patching.

**Fix:** disambiguate per row. E.g.:

```sql
UPDATE agent_connectors ac
SET label = CASE
        WHEN ac.is_default THEN c.name
        ELSE c.name || ' (' || substr(ac.connector_instance_id::text, 1, 8) || ')'
    END
FROM connectors c
WHERE c.id = ac.connector_id;
```

…or just drop the uniqueness requirement on the restored column (create a plain index) and add a note that rolling back loses user-set labels. Either is fine; what's not fine is a rollback that explodes.

---

frontend/src/pages/agents/connectors/ConnectorInstancesSection.tsx
```diff
@@ -269,40 +239,17 @@ function InstanceCard({
...
-    } catch (err) {
-      toast.error(err instanceof Error ? err.message : "Failed to rename.");
-      setLabelDraft(instance.label);
-      setEditing(false);
-    }
-  }
+  const isBusy = settingDefault || deleting;
+  const displayName =
+    instance.display?.trim() ||
+    `Instance ${instance.connector_instance_id.slice(0, 8)}…`;
```
chiedoclaude: **[Medium · UX]** After this PR, the "Add instance" flow creates an instance with no display name and shows `Instance 11111111…` until a credential is bound. The Add dialog copy ("Assign a credential below to set how it appears in approvals and capabilities.") points to a "below" that doesn't exist — the dialog only has Cancel / Add.

Two concrete issues for end users:

1. **Approval routing is unusable for a fresh instance.** The user's action needs `connector_instance: "<display>"`, but display is `""` until they remember to go back and bind a credential. The uniqueness issue flagged on `ResolveAgentConnectorInstance` makes this worse — the empty-display instance collides with the default instance (also empty display) and `resolve` picks one arbitrarily.

2. **UUID truncation label looks like a bug.** `Instance 11111111…` is indistinguishable from a placeholder skeleton.

**Fix suggestions (pick one):**

- Require credential assignment to happen in the same dialog (wizard flow: pick/create credential → create instance with that binding), so `display` is never empty.
- If you want to keep the two-step UX, at least render `Unnamed instance — assign a credential` and visually indicate it's incomplete, and update the dialog description to not reference UI that isn't there.

Either way, tracking "instance without a credential" as a first-class state is worth a follow-up.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 5 (1 critical, 3 high, 1 medium, 0 low)

Main themes, leading with severity:

1. **Build-breaking regression in `db/agent_connector_instances.go`** — the `"time"` import was removed but the struct still uses `time.Time`. The PR body explicitly notes `go test`/`go build` weren't run locally, so CI will be the first catch — but this is why those checks matter. Must fix.
2. **Concurrency / correctness regressions around the DB schema simplification.** Dropping `label` and its unique index surfaced two ordering-dependent bugs that didn't exist before:
   - `EnableAgentConnector` swapped `ON CONFLICT DO NOTHING` for a non-atomic `NOT EXISTS`; concurrent enables now race.
   - `ResolveAgentConnectorInstance` matches on a computed string that has no uniqueness guarantee, so `QueryRow` can silently route to the wrong instance.
   For an approval system, both of these are "data corruption in slow motion" risks.
3. **Down migration is broken** for any DB with multiple instances per (agent, connector) — it will fail on the unique-index recreation step. This needs to be fixable before merge; a rollback you can't execute isn't a rollback.
4. **One UX regression**: a freshly created instance has no way to display a usable name until a credential is bound, and the current fallback (`Instance 11111111…`) plus dialog copy that references non-existent UI makes this confusing.

### What Looks Good

- The OpenAPI spec and bundle are kept in lockstep with the schema change — including the omitempty/required-field tweaks, the capabilities description pointing at `display`, and the legacy-label note on `ApprovalSummary.action`. That's exactly the kind of parity that's usually missed.
- Good defense-in-depth on the frontend and mobile clients: both read `_connector_instance_display` with a principled fallback to the legacy `_connector_instance_label`, and `audit_helpers.go` preserves both in redacted JSON, so historical approvals keep rendering correctly.
- Tests were meaningfully upgraded, not just mechanically renamed — `TestAgentRequestApproval_MultiInstance_RequiresConnectorInstance` now actually seeds credentials on both instances and asserts the returned `available_instances` contains the expected credential labels. That's the right way to cover the derivation.
- Migration structure is clean: the BEFORE INSERT trigger is rewritten to drop the now-irrelevant label backfill but keeps the "first row is default" invariant. Nice.

---

---

chiedoclaude:
## Code Review Complete

**PR:** #979 — Phase 1: drop agent_connectors.label (migration + DB/API alignment)
**Author:** @chiedocursor
**Rounds completed:** 3/6
**Reason for stopping:** no activity for 10 minutes (HEAD unchanged at `fcf621b` across all rounds)

### Outstanding Issues

All 5 findings from Round 1 remain open:

1. **Critical — Build break.** `db/agent_connector_instances.go` removed `"time"` import but still uses `time.Time`.
2. **High — Correctness.** `EnableAgentConnector` race: `NOT EXISTS + INSERT` is not atomic after dropping the label unique index; concurrent enables can return unique-violation errors or split state.
3. **High — Correctness.** `ResolveAgentConnectorInstance` can silently route to the wrong instance because `COALESCE(cr.label, oc.extra_data->>'name', '')` isn't unique per `(agent, connector)` and `QueryRow` takes the first match.
4. **High — Migration safety.** The Down migration sets every restored row's `label` to `connectors.name`, then creates a unique index over it — breaks on any DB with >1 instance per `(agent, connector)`.
5. **Medium — UX.** New instances show `Instance 11111111…` until a credential is bound; add-dialog copy references UI that doesn't exist.

### Review History
| 1     | 5               | Initial review (1 critical, 3 high, 1 medium) |
| 2     | 0               | No new commits — clean round |
| 3     | 0               | No new commits — exited on inactivity timeout |

### Overall Assessment

The surface-level refactor is well executed — OpenAPI spec/bundle kept in sync, clients handle the legacy `_connector_instance_label` key gracefully, and the upgraded multi-instance tests actually exercise the new credential-derived `display` path. The Phase 2 client and mobile changes are particularly clean.

**But this PR isn't mergeable as-is.** It has a build-breaking Go compile error (missed because backend tests didn't run locally, as the PR body transparently calls out), a broken Down migration, and two new concurrency/ambiguity bugs that the old label-uniqueness invariant was silently preventing. The "instance display = credential nickname" idea is sound, but it needs an explicit uniqueness story per `(agent, connector)`; otherwise `connector_instance: "Engineering"` becomes a non-deterministic router.

Recommend: fix the compile error and Down migration before re-running CI, then decide on uniqueness semantics for the derived display (either enforce at credential-binding time, or switch `ResolveAgentConnectorInstance` to error on ambiguous matches).

---

---

frontend/src/pages/agents/connectors/ConnectorInstancesSection.tsx
chiedocursor: Fixed in 08037570: Add dialog copy now points to assigning a credential on the instance card in the list after closing the dialog; unnamed instances show "Unnamed instance — assign a credential" instead of UUID truncation.

db/agent_connectors.go
chiedocursor: Fixed in 08037570: `EnableAgentConnector` now uses `ON CONFLICT (agent_id, connector_id) WHERE (is_default) DO NOTHING` with the existing partial unique index, keeps the NOT EXISTS guard to avoid extra rows, and on unique-violation (23505) falls back to reading the existing row for idempotent success.

db/agent_connector_instances.go
chiedocursor: Fixed in 08037570: `ResolveAgentConnectorInstance` uses `Query` and returns `AgentConnectorInstanceErrAmbiguousDisplay` with `matching_instance_ids` in the API error details (`connector_instance_ambiguous`). Added `TestResolveAgentConnectorInstance_AmbiguousDisplay`.

db/agent_connector_instances.go
chiedocursor: Fixed in 08037570: restored `"time"` import for `EnabledAt time.Time`.

db/migrations/20260421134629_drop_agent_connectors_label.sql
chiedocursor: Fixed in 08037570: Down migration UPDATE now uses CASE to disambiguate non-default rows (`connector name (uuid-prefix)`) before recreating the unique index.

chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
- chiedoclaude asked for Restore time import for AgentConnectorInstance → implemented in `080375709e506714f973d2a59a0406e2b8df3876`
- chiedoclaude asked for EnableAgentConnector atomic idempotency with ON CONFLICT → implemented in `080375709e506714f973d2a59a0406e2b8df3876`
- chiedoclaude asked for ResolveAgentConnectorInstance ambiguous display → implemented in `080375709e506714f973d2a59a0406e2b8df3876`
- chiedoclaude asked for Migration Down disambiguate labels → implemented in `080375709e506714f973d2a59a0406e2b8df3876`
- chiedoclaude asked for Connector instances UX copy and unnamed label → implemented in `080375709e506714f973d2a59a0406e2b8df3876`
- chiedoclaude asked for Review Round 1 summary — addressed in 08037570 → implemented in `080375709e506714f973d2a59a0406e2b8df3876`
- chiedoclaude asked for Code Review Complete thread — findings addressed in 08037570 → implemented in `080375709e506714f973d2a59a0406e2b8df3876`

---
*Watch session ended after 0 minutes of inactivity. Processed 7 comments across 2 poll cycles.*

## 🔧 CI / audit remediation

### Remediation round 1

- **ci.yml** — conclusion `failure` — Backend Tests: duplicate TestEnableAgentConnector_Idempotent in agent_connectors_test.go vs agent_connectors_query_test.go and unused context import — merged idempotent test into query_test with row-count assertion; removed duplicate from schema test file. (`683d9d6f`)

### Remediation round 2

- **ci.yml** — conclusion `failure` — Backend Tests: unused import github.com/supersuit-tech/permission-slip/db in agent_connectors_test.go after moving idempotent test — removed import. (`3ff77115`)

### Remediation round 3

- **ci.yml** — conclusion `failure` — TestResolveAgentConnectorInstance_AmbiguousDisplay failed: cannot insert two credentials with same (user_id, service, label) — rewrote test to bind two OAuth connections with identical extra_data name instead. (`1882f5d7`)

### Remediation round 4

- **ci.yml** — conclusion `failure` — TestResolveAgentConnectorInstance_AmbiguousDisplay: InsertOAuthConnectionFull left scopes NULL — set Scopes: [] in opts for NOT NULL oauth_connections.scopes. (`0cb5479a`)

Issue #975, opened 2026-04-21
Show full email thread on approval screen for email reply actions (web + mobile)


chiedoclaude:
## Problem

When an agent replies to an email via the Google or Microsoft connector, the approval screen (web and mobile) only shows the reply body the agent is about to send. The reviewer has no context about **what** is being replied to — they can't see the original sender, the subject, the message body, or any prior messages in the thread.

This makes it difficult for a human approver to sanity-check a reply. They have to leave the approval UI, open their inbox, find the thread, and read it themselves before they feel comfortable clicking "Approve".

We should surface the **full conversation history** of the email thread directly on the approval screen so the reviewer can make an informed decision without leaving the app.

## Scope

- Applies to the `google.send_email_reply` action (already exists) and a new `microsoft.send_email_reply` action (does not yet exist — Microsoft only has `send_email` today).
- Applies to **both** the web approval dialog (`ReviewApprovalDialog.tsx`) and the mobile approval detail screen (`ApprovalDetailScreen.tsx`).
- Full conversation history (all prior messages in the thread), not just the single message being replied to.
- Attachments are out of scope for this issue — see follow-up issue for the full attachment gap. For now, we only list the filename so the reviewer knows an attachment exists.

## Design decisions (already resolved)

- **Thread scope:** full conversation history.
- **HTML rendering:** sanitized HTML on web (sandboxed iframe or DOMPurify); plain text on mobile (simpler, avoids RN HTML renderer footguns).
- **Microsoft reply:** dedicated `microsoft.send_email_reply` action using Graph `POST /me/messages/{id}/reply`, for parity with Google.
- **Attachments:** list filename only in this issue; deeper handling tracked separately.
- **Size cap:** truncate each message body at ~20 KB server-side with a `truncated: true` flag; UI offers a "show more" affordance.

## Shipping plan (PRs)

This issue ships as **three PRs**, not one per phase. Phase 4 is manual verification applied across PRs 2 and 3.

1. **PR 1 — Backend (Phase 1, Google + Microsoft + OpenAPI together).** The Google thread fetch, new Microsoft reply action, and OpenAPI `email_thread` schema all land in one PR. They share the schema definition, so keeping them together avoids a conflict and ships a complete backend contract. Safe to merge without UI — the new field just sits unused until the frontends consume it.
2. **PR 2 — Web UI (Phase 2).** Depends on PR 1 being merged.
[...]

PR #980, opened 2026-04-21
Email replies Phase 1: backend email_thread + Microsoft send_email_reply

chiedocursor:
## Summary

- **Gmail:** `ResolveResourceDetails` for `google.send_email_reply` now loads the full thread via `GET .../threads/{id}?format=full`, parses MIME bodies (including multipart), lists attachment filenames/sizes, applies a ~20 KB per-field truncation flag, and exposes `email_thread` in resolved details.
- **Microsoft:** New `microsoft.send_email_reply` action (`POST /me/messages/{id}/reply`), manifest entry, template, and `ResolveResourceDetails` that fetches the anchor message plus the conversation by `conversationId` and maps it to the same normalized shape (HTML → plain text via a shared stripper).
- **API:** When creating an approval, if resolved `resource_details` contains `email_thread`, it is merged into `context.details.email_thread` (skipped if the merged JSON would exceed the existing 64 KiB context limit). OpenAPI documents `EmailThread` / `EmailThreadMessage` / `EmailThreadAttachment` under `ActionContext.details`.

Part of #975

### Claude Code

- [x] `make bundle` — OpenAPI bundle succeeds (run locally).
- [x] `cd frontend && npm run generate:api` — regenerates types from bundle (schema.d.ts is gitignored).
- [x] **Go tests not run in this environment:** local `go test` fails because `go.mod` requires Go 1.25.9 while the sandbox has an older toolchain. CI should run `make test-backend` / full `make test`.

### Human

- [ ] After merge, verify a pending approval for `google.send_email_reply` / `microsoft.send_email_reply` includes `context.details.email_thread` in API responses.
- [ ] Smoke-test Microsoft reply against a real mailbox (Graph permissions already include Mail.Send/Mail.Read).

https://claude.ai/code

<div><a href="https://cursor.com/agents/bc-ea91cab9-f900-4922-9bb0-37cf89203724"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ea91cab9-f900-4922-9bb0-37cf89203724"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/api/approval_email_thread.go
+++ b/api/approval_email_thread.go
@@ -1,0 +1,59 @@
+package api
+
+import (
+	"encoding/json"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+)
+
+// mergeEmailThreadFromResourceDetailsIntoContext copies email_thread from resolved
+// resource_details into context.details.email_thread for approval UI (issue #975).
+// If resource_details is nil or has no email_thread, returns contextJSON unchanged.
+func mergeEmailThreadFromResourceDetailsIntoContext(contextJSON json.RawMessage, resourceDetails []byte) json.RawMessage {
+	if len(resourceDetails) == 0 {
+		return contextJSON
+	}
+	var rd map[string]json.RawMessage
+	if err := json.Unmarshal(resourceDetails, &rd); err != nil {
+		return contextJSON
+	}
+	threadRaw, ok := rd["email_thread"]
+	if !ok || len(threadRaw) == 0 {
+		return contextJSON
+	}
+	var thread connectors.EmailThreadPayload
+	if err := json.Unmarshal(threadRaw, &thread); err != nil {
+		return contextJSON
+	}
+
+	var ctxObj map[string]json.RawMessage
+	if err := json.Unmarshal(contextJSON, &ctxObj); err != nil || ctxObj == nil {
+		return contextJSON
+	}
+	detailsRaw, hasDetails := ctxObj["details"]
+	var detailsObj map[string]json.RawMessage
+	if hasDetails && len(detailsRaw) > 0 {
+		_ = json.Unmarshal(detailsRaw, &detailsObj)
+	}
+	if detailsObj == nil {
+		detailsObj = map[string]json.RawMessage{}
+	}
+	if _, exists := detailsObj["email_thread"]; exists {
+		return contextJSON
+	}
+	detailsObj["email_thread"] = threadRaw
+	mergedDetails, err := json.Marshal(detailsObj)
+	if err != nil {
+		return contextJSON
+	}
+	ctxObj["details"] = mergedDetails
+	out, err := json.Marshal(ctxObj)
+	if err != nil {
+		return contextJSON
+	}
+	// Preserve the 65536-byte cap from the request validator.
+	if len(out) > 65536 {
+		return contextJSON
+	}
+	return out
+}
--- a/connectors/google/gmail_thread.go
+++ b/connectors/google/gmail_thread.go
@@ -143,12 +143,6 @@ func extractHTMLAndPlain(part *gmailMessagePart, depth int) (plain, html string)
 		return plainOut, htmlOut
 	}
 
-	if part.Body.Data != "" && part.MimeType == "text/plain" {
-		return decodeBase64URL(part.Body.Data), ""
-	}
-	if part.Body.Data != "" && part.MimeType == "text/html" {
-		return "", decodeBase64URL(part.Body.Data)
-	}
 	return "", ""
 }
 
--- a/connectors/google/gmail_thread.go
+++ b/connectors/google/gmail_thread.go
@@ -1,0 +1,171 @@
+package google
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+)
+
+// gmailThreadAPIResponse is the Gmail threads.get API response (format=full).
+type gmailThreadAPIResponse struct {
+	ID       string             `json:"id"`
+	Messages []gmailFullMessage `json:"messages"`
+}
+
+// buildGmailEmailThread fetches the full thread and returns a normalized email_thread payload.
+func (c *GoogleConnector) buildGmailEmailThread(ctx context.Context, creds connectors.Credentials, threadID string) (*connectors.EmailThreadPayload, error) {
+	if threadID == "" {
+		return nil, fmt.Errorf("missing thread_id")
+	}
+	threadURL := c.gmailBaseURL + "/gmail/v1/users/me/threads/" + url.PathEscape(threadID) + "?format=full"
+	var thr gmailThreadAPIResponse
+	if err := c.doJSON(ctx, creds, http.MethodGet, threadURL, nil, &thr); err != nil {
+		return nil, err
+	}
+	if len(thr.Messages) == 0 {
+		return &connectors.EmailThreadPayload{Subject: "", Messages: nil}, nil
+	}
+
+	// Oldest → newest by internalDate (fallback: preserve API order).
+	msgs := append([]gmailFullMessage(nil), thr.Messages...)
+	sort.Slice(msgs, func(i, j int) bool {
+		ai, _ := strconv.ParseInt(msgs[i].InternalDate, 10, 64)
+		aj, _ := strconv.ParseInt(msgs[j].InternalDate, 10, 64)
+		return ai < aj
+	})
+
+	subject := ""
+	for _, m := range msgs {
+		if s := headerValue(&m.Payload, "Subject"); s != "" {
+			subject = stripHeaderNewlines(s)
+			break
+		}
+	}
+
+	out := make([]connectors.EmailThreadMessage, 0, len(msgs))
+	for _, m := range msgs {
+		em := connectors.EmailThreadMessage{
+			From:      stripHeaderNewlines(headerValue(&m.Payload, "From")),
+			To:        splitAddressList(headerValue(&m.Payload, "To")),
+			Cc:        splitAddressList(headerValue(&m.Payload, "Cc")),
+			Date:      gmailMessageDateRFC3339(&m),
+			Snippet:   m.Snippet,
+			MessageID: m.ID,
+		}
+		plain, html := extractHTMLAndPlain(&m.Payload, 0)
+		em.BodyText = plain
+		em.BodyHTML = html
+		if em.BodyText == "" && em.BodyHTML != "" {
+			em.BodyText = htmlToPlainText(em.BodyHTML)
+		}
+		em.Attachments = gmailAttachmentsToThread(extractAttachments(&m.Payload))
+		connectors.TruncateEmailThreadBodies(&em)
+		out = append(out, em)
+	}
+
+	return &connectors.EmailThreadPayload{
+		Subject:  subject,
+		Messages: out,
+	}, nil
+}
+
+func headerValue(part *gmailMessagePart, name string) string {
+	if part == nil {
+		return ""
+	}
+	lower := strings.ToLower(name)
+	for _, h := range part.Headers {
+		if strings.ToLower(h.Name) == lower {
+			return h.Value
+		}
+	}
+	return ""
+}
+
+func splitAddressList(s string) []string {
+	if strings.TrimSpace(s) == "" {
+		return nil
+	}
+	parts := strings.Split(s, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p != "" {
+			out = append(out, stripHeaderNewlines(p))
+		}
+	}
+	return out
+}
+
+func gmailMessageDateRFC3339(m *gmailFullMessage) string {
+	if m.InternalDate != "" {
+		if ms, err := strconv.ParseInt(m.InternalDate, 10, 64); err == nil {
+			return time.UnixMilli(ms).UTC().Format(time.RFC3339)
+		}
+	}
+	if d := headerValue(&m.Payload, "Date"); d != "" {
+		return stripHeaderNewlines(d)
+	}
+	return ""
+}
+
+// extractHTMLAndPlain walks the MIME tree and returns best-effort text/plain and text/html bodies.
+func extractHTMLAndPlain(part *gmailMessagePart, depth int) (plain, html string) {
+	if part == nil || depth > maxMIMEDepth {
+		return "", ""
+	}
+
+	if part.Body.Data != "" && part.MimeType == "text/plain" {
+		return decodeBase64URL(part.Body.Data), ""
+	}
+	if part.Body.Data != "" && part.MimeType == "text/html" {
+		return "", decodeBase64URL(part.Body.Data)
+	}
+
+	if strings.HasPrefix(part.MimeType, "multipart/") {
+		var plainOut, htmlOut string
+		for i := range part.Parts {
+			p, h := extractHTMLAndPlain(&part.Parts[i], depth+1)
+			if p != "" {
+				plainOut = p
+			}
+			if h != "" {
+				htmlOut = h
+			}
+		}
+		return plainOut, htmlOut
+	}
+
+	if part.Body.Data != "" && part.MimeType == "text/plain" {
+		return decodeBase64URL(part.Body.Data), ""
+	}
+	if part.Body.Data != "" && part.MimeType == "text/html" {
+		return "", decodeBase64URL(part.Body.Data)
+	}
+	return "", ""
+}
+
+func gmailAttachmentsToThread(in []gmailAttachmentInfo) []connectors.EmailThreadAttachment {
+	if len(in) == 0 {
+		return nil
+	}
+	out := make([]connectors.EmailThreadAttachment, 0, len(in))
+	for _, a := range in {
+		fn := a.Filename
+		if fn == "" {
+			fn = "(no filename)"
+		}
+		out = append(out, connectors.EmailThreadAttachment{
+			Filename:  fn,
+			SizeBytes: int64(a.Size),
+		})
+	}
+	return out
+}
--- a/connectors/google/read_email.go
+++ b/connectors/google/read_email.go
@@ -73,6 +73,7 @@ type gmailFullMessage struct {
 type gmailMessagePart struct {
 	PartID   string `json:"partId"`
 	MimeType string `json:"mimeType"`
+	Filename string `json:"filename"`
 	Headers  []struct {
 		Name  string `json:"name"`
 		Value string `json:"value"`
@@ -87,17 +88,17 @@ type gmailMessagePart struct {
 
 // emailFullDetail is the shape returned to the agent.
 type emailFullDetail struct {
-	ID          string               `json:"id"`
-	ThreadID    string               `json:"thread_id"`
-	From        string               `json:"from,omitempty"`
-	To          string               `json:"to,omitempty"`
-	Cc          string               `json:"cc,omitempty"`
-	Subject     string               `json:"subject,omitempty"`
-	Date        string               `json:"date,omitempty"`
-	Snippet     string               `json:"snippet,omitempty"`
-	Labels      []string             `json:"labels,omitempty"`
-	ContentType string               `json:"content_type,omitempty"`
-	Body        string               `json:"body,omitempty"`
+	ID          string                `json:"id"`
+	ThreadID    string                `json:"thread_id"`
+	From        string                `json:"from,omitempty"`
+	To          string                `json:"to,omitempty"`
+	Cc          string                `json:"cc,omitempty"`
+	Subject     string                `json:"subject,omitempty"`
+	Date        string                `json:"date,omitempty"`
+	Snippet     string                `json:"snippet,omitempty"`
+	Labels      []string              `json:"labels,omitempty"`
+	ContentType string                `json:"content_type,omitempty"`
+	Body        string                `json:"body,omitempty"`
 	Attachments []gmailAttachmentInfo `json:"attachments,omitempty"`
 }
 
@@ -244,6 +245,9 @@ func collectAttachments(part *gmailMessagePart, out *[]gmailAttachmentInfo, dept
 				}
 			}
 		}
+		if info.Filename == "" && part.Filename != "" {
+			info.Filename = part.Filename
+		}
 		*out = append(*out, info)
 	}
 	for i := range part.Parts {
@@ -282,7 +286,7 @@ func parseFilename(headerValue string) string {
 	return plainName
 }
 
-// decodeRFC5987 decodes a RFC 5987 encoded value like "UTF-8''caf%C3%A9.pdf".
+// decodeRFC5987 decodes a RFC 5987 encoded value like "UTF-8”caf%C3%A9.pdf".
 // Only UTF-8 charset is supported; other charsets (e.g. ISO-8859-1) return ""
 // to avoid producing invalid UTF-8 strings.
 func decodeRFC5987(value string) string {
@@ -309,7 +313,10 @@ func decodeBase64URL(s string) string {
 	if err != nil {
 		b, err = base64.URLEncoding.DecodeString(s)
 		if err != nil {
-			return s // return as-is if decoding fails
+			b, err = base64.StdEncoding.DecodeString(s)
+			if err != nil {
+				return s // return as-is if decoding fails
+			}
 		}
 	}
 	return string(b)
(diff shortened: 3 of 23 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### PR Checklist Items
- ✅ **``make bundle` — OpenAPI bundle succeeds (run locally).`** — Ran `make bundle`; Redocly bundled spec/openapi/openapi.yaml to openapi.bundle.yaml successfully. (``)
- ✅ **``cd frontend && npm run generate:api` — regenerates types from bundle (schema.d.ts is gitignored).`** — Ran `npm run generate:api` from frontend; openapi-typescript updated gitignored schema.d.ts. (``)
- ✅ **`**Go tests not run in this environment:** local `go test` fails because `go.mod` requires Go 1.25.9 while the sandbox has an older toolchain. CI should run `make test-backend` / full `make test`.`** — Confirmed: `go test` cannot run locally here (toolchain mismatch). Reliance on CI documented in checklist. (``)

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

## 🔧 CI / audit remediation

### Remediation round 1

- **ci.yml** — conclusion `failure` — Backend tests failed: (1) TestAllActionsHaveDisplayTemplate — microsoft had only send_email_reply with DisplayTemplate while other actions had none; removed DisplayTemplate from microsoft.send_email_reply. (2) TestBuildGmailEmailThread_MultipartAndAttachments — Gmail API uses standard Base64 for body.data in threads; added StdEncoding fallback in decodeBase64URL, use MIME part filename for attachments, simplified extractHTMLAndPlain duplicate branches. (`20a6c51c`)

connectors/microsoft/microsoft.go
```diff
@@ -109,6 +109,32 @@ func (c *MicrosoftConnector) Manifest() *connectors.ConnectorManifest {
 					}
 				}`)),
 			},
+			{
+				ActionType:      "microsoft.send_email_reply",
+				Name:            "Reply to Email",
+				Description:     "Reply to an existing message in a Microsoft 365 mailbox",
+				RiskLevel:       "medium",
+				DisplayTemplate: "Reply to email {{message_id}}",
```
chiedoclaude: **[Critical · Correctness]** This is the first Microsoft action to get a `DisplayTemplate`. `TestAllActionsHaveDisplayTemplate` in `connectors/display_template_test.go:97-124` enforces that **once a connector defines any templates, ALL of its actions must have one** — this will now fail for the other 24 Microsoft actions (`send_email`, `list_emails`, `create_calendar_event`, `list_drive_files`, …) which all have empty `DisplayTemplate`.

**Why:** CI will go red on `make test-backend`. Either:
- Add `DisplayTemplate` strings to every Microsoft action in the same PR (preferred, consistent with Google), or
- Drop this template for now and defer DisplayTemplate adoption to a dedicated PR that does them all.

Also worth noting: `"Reply to email {{message_id}}"` renders the raw opaque Graph message ID to the approver, which isn't very human-friendly. If you keep it, consider referencing the resolved `{{subject}}` (populated by `resolveSendEmailReply`) like Google's `"Send email to {{to}} — {{subject}}"` pattern — just remember `subject` and `from` must be in `resourceDetailFields` in `display_template_test.go` for the parameter-existence check to pass.

---

connectors/google/gmail_thread.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		return plainOut, htmlOut
+	}
+
+	if part.Body.Data != "" && part.MimeType == "text/plain" {
+		return decodeBase64URL(part.Body.Data), ""
+	}
+	if part.Body.Data != "" && part.MimeType == "text/html" {
+		return "", decodeBase64URL(part.Body.Data)
+	}
```
chiedoclaude: **[High · Correctness]** Dead code: lines 146-151 duplicate the two early-return branches at lines 125-130 verbatim. They're unreachable — if `part.MimeType == "text/plain"`/`"text/html"` with `Body.Data != ""`, the function already returned at line 125 or 128. If it didn't return there, it's because the MIME type didn't match, so repeating the exact same check at the end can never succeed.

**Why:** Unreachable code suggests intent may diverge from behavior (e.g. maybe you meant to handle a sub-part case here?). Either delete these lines or replace them with the intended fallback logic.

**Fix:**
```go
if strings.HasPrefix(part.MimeType, "multipart/") {
    var plainOut, htmlOut string
    for i := range part.Parts {
        p, h := extractHTMLAndPlain(&part.Parts[i], depth+1)
        if p != "" { plainOut = p }
        if h != "" { htmlOut = h }
    }
    return plainOut, htmlOut
}
return "", ""
```

---

api/approval_email_thread.go
```diff
@@ -0,0 +1,59 @@
...
+		return contextJSON
+	}
+	ctxObj["details"] = mergedDetails
+	out, err := json.Marshal(ctxObj)
+	if err != nil {
+		return contextJSON
+	}
+	// Preserve the 65536-byte cap from the request validator.
+	if len(out) > 65536 {
+		return contextJSON
```
chiedoclaude, line 56: **[High · Observability]** When the merged context exceeds 64 KiB, the email_thread is silently dropped — no log, no metric, no indication on the approval. An approver opening a long reply thread will see only the basic `subject`/`from` from `resource_details` and have no way to know that the backend *could* have populated a thread but didn't.

**Why:** Long threads are exactly the cases where the approval UI most needs the context. Silent drops make this impossible to diagnose in production ("why does this approval show no thread?").

**Fix:** At minimum `log.Printf` with the approval's action type and the overflow size so it surfaces in searches. Better: trim message bodies further (e.g. drop oldest messages or halve `MaxEmailThreadBodyRunes` on overflow) before giving up. A `"truncated": true` flag at the thread level would also let the UI warn approvers that they're seeing a partial view.

---

connectors/google/resolve_resource_details.go
```diff
@@ -232,12 +232,34 @@ func (c *GoogleConnector) resolveEmail(ctx context.Context, creds connectors.Cre
...
+	meta, err := c.fetchEmailMetadata(ctx, creds, p.MessageID)
+	if err != nil {
+		return nil, err
+	}
+	if p.ThreadID == "" {
+		return meta, nil
+	}
+	thread, err := c.buildGmailEmailThread(ctx, creds, p.ThreadID)
+	if err != nil {
+		return meta, nil // non-fatal: keep subject/from for display template
```
chiedoclaude, line 250: **[High · Observability]** `buildGmailEmailThread` errors are discarded. The comment says "non-fatal" which is correct — we don't want to block the approval — but we shouldn't throw away the error entirely.

**Why:** When thread fetch fails (Gmail quota, missing scope, deleted thread, transient 5xx), the approval is stored without a thread and there is zero trace in logs or Sentry. Debugging "approvals missing email_thread" requires a trail.

**Fix:** Match the pattern already used in `api/agent_approvals.go:295-296` where `ResolveResourceDetails` errors are logged + captured:
```go
thread, err := c.buildGmailEmailThread(ctx, creds, p.ThreadID)
if err != nil {
    log.Printf("resolveEmailReply: buildGmailEmailThread(%q): %v", p.ThreadID, err)
    return meta, nil
}
```
(Or simply bubble the error up — the outer caller in `handleAgentRequestApproval` already logs + captures and falls back gracefully.)

---

connectors/microsoft/resolve_resource_details.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	details := map[string]any{
+		"subject": stripGraphHeaderNewlines(msg.Subject),
+	}
+	if msg.From != nil {
+		details["from"] = stripGraphHeaderNewlines(msg.From.EmailAddress.Address)
+	}
+	thread, err := c.buildMicrosoftEmailThread(ctx, creds, p.MessageID)
+	if err != nil {
+		return details, nil
```
chiedoclaude, line 39: **[High · Observability]** Same issue as the Gmail resolver: `buildMicrosoftEmailThread` errors are silently swallowed.

Additionally, the code inside `buildMicrosoftEmailThread` silently falls back to `singleMessageThread(src)` when the conversation listing fails (`thread.go:93-95`), so there are actually **two layers** of silent failure here. A Graph throttling event or permission issue on `$filter=conversationId` will quietly downgrade to a single-message thread, and then *that* returning no error means there's no signal anywhere.

**Fix:** Log at the resolver boundary, and also log inside `buildMicrosoftEmailThread` before the fallback to single-message thread. At minimum `log.Printf` so operators can correlate missing threads with Graph errors.

---

connectors/microsoft/thread.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	filter := fmt.Sprintf("conversationId eq '%s'", escapeODataString(convID))
+	order := "receivedDateTime asc"
+	selectFields := "id,subject,conversationId,receivedDateTime,bodyPreview,from,toRecipients,ccRecipients,body"
+	path := "/me/messages?$filter=" + url.QueryEscape(filter) +
+		"&$orderby=" + url.QueryEscape(order) +
+		"&$select=" + url.QueryEscape(selectFields) +
+		"&$expand=" + url.QueryEscape("attachments($select=name,size)") +
+		"&$top=100"
```
chiedoclaude, line 90: **[Medium · Data completeness]** `$top=100` without following `@odata.nextLink`. For a long-running email conversation (support tickets, mailing-list threads), messages beyond the first 100 are silently dropped — the approver sees a truncated thread with no indication.

**Why:** `receivedDateTime asc` + `$top=100` returns the *oldest* 100, so the most recent (and typically most relevant) messages are the ones being lost. That's the opposite of what you want for reply context.

**Fix options (pick one):**
1. Order `receivedDateTime desc`, take 100, reverse the slice so the UI still sees oldest→newest but prioritizes recent context.
2. Follow `@odata.nextLink` up to a safe cap (e.g. 500 messages).
3. Set a `truncated` flag on `EmailThreadPayload` when `@odata.nextLink != ""` so the UI can warn approvers that the thread is partial.

---

connectors/html_plain.go
```diff
@@ -0,0 +1,21 @@
...
+func StripHTMLToPlain(s string) string {
+	s = strings.ReplaceAll(s, "<br>", "\n")
+	s = strings.ReplaceAll(s, "<br/>", "\n")
+	s = strings.ReplaceAll(s, "<br />", "\n")
+	s = strings.ReplaceAll(s, "</p>", "\n")
+	s = strings.ReplaceAll(s, "</div>", "\n")
+	s = htmlTagPattern.ReplaceAllString(s, "")
+	s = html.UnescapeString(s)
+	return strings.TrimSpace(s)
+}
```
chiedoclaude, line 21: **[Medium · DRY]** `StripHTMLToPlain` is a near-exact duplicate of `htmlToPlainText` in `connectors/google/email_helpers.go:22-33` (same tag-replacement list, same regex, same `html.UnescapeString`, same `TrimSpace`). `gmail_thread.go:66` calls the Google-local version, while `microsoft/thread.go:146` calls the shared one — so the two providers now go through nominally-different code paths that happen to produce the same output.

**Why:** Inevitably these will drift. If someone fixes a bug (e.g. `<br/>` variants, `&nbsp;` handling) in one, the other silently retains the old behavior.

**Fix:** Keep the shared `connectors.StripHTMLToPlain`, delete the Google-local `htmlToPlainText` + `tagPattern`, and update the two call sites (`gmail_thread.go:66` and `email_helpers.go:64`) to use `connectors.StripHTMLToPlain`.

---

connectors/microsoft/resolve_resource_details.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return nil, err
+	}
+	details := map[string]any{
+		"subject": stripGraphHeaderNewlines(msg.Subject),
+	}
+	if msg.From != nil {
+		details["from"] = stripGraphHeaderNewlines(msg.From.EmailAddress.Address)
+	}
+	thread, err := c.buildMicrosoftEmailThread(ctx, creds, p.MessageID)
```
chiedoclaude, line 37: **[Medium · Performance]** Redundant Graph API call. `fetchGraphMessage(p.MessageID)` was just called on line 27, and then `buildMicrosoftEmailThread(p.MessageID)` calls `fetchGraphMessage(p.MessageID)` *again* internally (`thread.go:74`) with the same `$select`/`$expand` fields. That's two `GET /me/messages/{id}` per approval.

**Why:** Each approval submission now does 3 Graph calls instead of 2 (anchor msg ×2 + conversation listing). Microsoft Graph has per-app and per-tenant throttling; halving the anchor fetch is easy savings and also shortens the 5-second timeout budget the resolver runs under (`api/agent_approvals.go:291`).

**Fix:** Refactor `buildMicrosoftEmailThread` to accept an already-loaded `*graphFullMessage`, or split it into `buildMicrosoftEmailThreadFromMessage(msg *graphFullMessage)` so the resolver can share the single fetch:
```go
msg, err := c.fetchGraphMessage(ctx, creds, p.MessageID)
if err != nil { return nil, err }
// build details from msg
thread, _ := c.buildMicrosoftEmailThreadFromMessage(ctx, creds, msg)
```

---

connectors/google/gmail_thread.go
```diff
@@ -0,0 +1,171 @@
...
+		return &connectors.EmailThreadPayload{Subject: "", Messages: nil}, nil
+	}
+
+	// Oldest → newest by internalDate (fallback: preserve API order).
+	msgs := append([]gmailFullMessage(nil), thr.Messages...)
+	sort.Slice(msgs, func(i, j int) bool {
+		ai, _ := strconv.ParseInt(msgs[i].InternalDate, 10, 64)
+		aj, _ := strconv.ParseInt(msgs[j].InternalDate, 10, 64)
+		return ai < aj
+	})
```
chiedoclaude, line 42: **[Medium · Correctness]** `sort.Slice` is not stable. When two messages share the same `internalDate` (auto-reply vs. reply sent in the same millisecond, programmatic sends, bulk-imported threads), output order is nondeterministic — each rebuild of the same thread can produce a different ordering.

**Why:** Nondeterminism in approval context is a debugging pain: the stored approval, audit logs, and approver UI can disagree across re-renders. Tests against real data may also flake.

**Fix:** `sort.SliceStable(msgs, …)` — same signature, same O(n log n), and preserves API order as the tiebreaker.

---

connectors/microsoft/send_email_reply.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	path := "/me/messages/" + url.PathEscape(params.MessageID) + "/reply"
+	ct := graphBodyContentType(params.HTML)
+	var body graphReplyRequest
+	if ct == "Text" {
+		body.Comment = params.Body
+	} else {
+		body.Message.Body = graphEmailBody{ContentType: "HTML", Content: params.Body}
+	}
```
chiedoclaude, line 68: **[Low · Style]** Comparing `graphBodyContentType(params.HTML)` to the string literal `"Text"` is brittle — if `graphBodyContentType` ever adds a new return value or changes casing, this branches incorrectly with no compile-time error.

**Fix:** Branch on the source flag directly:
```go
if params.HTML != nil && !*params.HTML {
    body.Comment = params.Body
} else {
    body.Message.Body = graphEmailBody{ContentType: "HTML", Content: params.Body}
}
```
Or, if you want to keep the helper, introduce exported `GraphContentTypeText`/`GraphContentTypeHTML` constants instead of the literal.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 10 (1 critical, 4 high, 4 medium, 1 low)

This is a thoughtful, well-structured feature — the shared `EmailThreadPayload` schema + per-provider builders keep the two code paths consistent, the 20 KiB per-field truncation is sensible, and the tests cover the multipart/truncation happy paths nicely.

The one critical item is a test-breaking side-effect of adding the first Microsoft `DisplayTemplate` — `TestAllActionsHaveDisplayTemplate` enforces "once a connector defines any templates, all its actions must have one," so the other 24 Microsoft actions will now fail the test. Worth fixing before merge.

Most other findings cluster around **observability gaps** — thread-fetch failures and 64 KiB-overflow drops are all silent, so in production a missing `email_thread` in the approval UI will have no diagnostic trail. A second cluster is minor **code-hygiene** items (dead code in `extractHTMLAndPlain`, a duplicated `StripHTMLToPlain`/`htmlToPlainText` pair, a redundant Graph call, and non-stable sort).

### What Looks Good
- Shared `EmailThreadPayload` + `TruncateEmailThreadBodies` keep Gmail/Microsoft output consistent.
- `EmailThreadDetailsMap` round-trip through `json.Marshal`/`Unmarshal` avoids leaking Go types into the `map[string]any` payload.
- Header sanitization (`stripHeaderNewlines` / `stripGraphHeaderNewlines`) is applied consistently to all Gmail-sourced string fields.
- The OpenAPI schema additions (`EmailThread`, `EmailThreadMessage`, `EmailThreadAttachment`) are complete with `required` fields marked — should give good codegen.
- The 64 KiB cap check in `mergeEmailThreadFromResourceDetailsIntoContext` preserves the request validator's invariant correctly.
- Good test coverage of the truncation boundary in both Gmail and Microsoft builders.

---

---

connectors/google/read_email.go
```diff
@@ -282,7 +286,7 @@ func parseFilename(headerValue string) string {
 	return plainName
 }
 
-// decodeRFC5987 decodes a RFC 5987 encoded value like "UTF-8''caf%C3%A9.pdf".
+// decodeRFC5987 decodes a RFC 5987 encoded value like "UTF-8”caf%C3%A9.pdf".
```
chiedoclaude, line 289: **[Low · Documentation]** The docstring example got mangled by the fix commit — the two ASCII single quotes (`''`) in `UTF-8''caf%C3%A9.pdf` were replaced with a single Unicode right double quotation mark (U+201D, `”`), so the comment now reads `"UTF-8”caf%C3%A9.pdf"` and no longer matches what RFC 5987 actually specifies.

**Why:** The code still works (it splits on ASCII `'` in `strings.SplitN(value, "'", 3)`), but the comment is now actively misleading — a future reader looking up RFC 5987 will see the curly quote and assume the encoding uses a different separator. Cosmetic but worth a one-char fix since it's right in the docstring for the function.

**Fix:**
```go
// decodeRFC5987 decodes a RFC 5987 encoded value like "UTF-8''caf%C3%A9.pdf".
```
(restore the two straight ASCII single quotes)

---

chiedoclaude (commented):
## Review Round 2/6

**Findings:** 1 new (1 low) + 8 still-open from Round 1 (3 high, 4 medium, 1 low)

### Changes Since Last Round
Fix commit `20a6c51` ("fix(connectors): CI — Gmail base64 + attachment filename; drop partial MS display template") addressed **2 of 10** Round 1 findings:

- ✅ **Critical · Architecture** — Removed the partial `DisplayTemplate` from `microsoft.send_email_reply` so `TestAllActionsHaveDisplayTemplate` stays green.
- ✅ **Low · Code Quality** — Removed the duplicated leaf branches at the end of `extractHTMLAndPlain`; the function now falls through cleanly to `return "", ""` after the multipart walk.

Nice additional improvements that weren't explicitly requested:
- `decodeBase64URL` now falls back to standard base64 after URL-safe variants fail — a reasonable belt-and-braces addition for legacy Gmail payloads.
- `collectAttachments` now honors `gmailMessagePart.Filename` when the Content-Disposition header omits the filename — fixes a real bug where inline attachments would lose their name.

### Remaining Issues (from Round 1, not yet addressed)

**High**
- `api/approval_email_thread.go:54-56` — the 64 KiB overflow path silently returns the unmerged context. In production a missing `email_thread` in the approval UI will have no diagnostic trail. At minimum log a warning with approval_id + merged size.
- `connectors/google/resolve_resource_details.go:250` — Gmail thread fetch errors swallowed with `return meta, nil` and no log. Same observability concern.
- `connectors/microsoft/resolve_resource_details.go:39` — Microsoft thread fetch errors swallowed similarly, and additionally `buildMicrosoftEmailThread` itself swallows the `$filter` conversation query failure (`thread.go`, the `return singleMessageThread(src), nil` on `doRequest` error). Two layers of silent fallback.
- `connectors/microsoft/thread.go:90` — `$top=100` with `receivedDateTime asc` truncates to the **oldest** 100 messages when a conversation has more, so the most recent replies (the ones the agent is actually replying to) get dropped. Either switch to `desc` + reverse, or follow `@odata.nextLink` to pull the whole conversation.

**Medium**
- `connectors/html_plain.go` vs `connectors/google/email_helpers.go:22-33` — `StripHTMLToPlain` and `htmlToPlainText` are near-identical; consolidate to one shared implementation.
- `connectors/microsoft/resolve_resource_details.go:27` + `connectors/microsoft/thread.go` — `fetchGraphMessage` is called twice for the same `message_id` (once at line 27, once inside `buildMicrosoftEmailThread`). Pass the already-fetched `msg` into the thread builder or skip the outer fetch entirely.
- `connectors/google/gmail_thread.go:42` — `sort.Slice` is not stable; when two messages share an `internalDate` (rare but possible for bulk-imported threads) their order is non-deterministic. Use `sort.SliceStable`.
- `connectors/microsoft/send_email_reply.go:64` — `if ct == "Text"` compares against a string literal returned from `graphBodyContentType`. A typo anywhere in that helper silently sends HTML as the `message.body` payload. Use a typed constant or the boolean `*params.HTML` check directly.

### New in This Round
- **Low · Documentation** — The `decodeRFC5987` docstring got a cosmetic regression from the fix commit: `"UTF-8''caf%C3%A9.pdf"` (two ASCII single quotes) became `"UTF-8”caf%C3%A9.pdf"` (one U+201D right double quote). Code is still correct; just the comment is now misleading. Inline comment posted.

### What Looks Good
- Fast turnaround on the CI-breaker — removing the partial `DisplayTemplate` was the right call.
- The base64 fallback and `gmailMessagePart.Filename` fix were nice opportunistic improvements beyond the reported issues.
- The dead-code cleanup in `extractHTMLAndPlain` shows the fix was thought through rather than just silencing the linter.

---

---

chiedoclaude:
## Code Review Complete

**PR:** #980 — Email replies Phase 1: backend email_thread + Microsoft send_email_reply
**Author:** @chiedocursor
**Rounds completed:** 2/6
**Reason for stopping:** PR merged before Round 3

### Outstanding Issues
The PR merged with 8 Round 1 findings still unaddressed. Most are non-blocking but worth a follow-up PR since several affect production behavior. Listed in priority order:

**High — observability & correctness**
- `api/approval_email_thread.go:54-56` — silent drop when merged context > 64 KiB. Add a structured log with approval_id and merged size so missing `email_thread` in the UI is diagnosable.
- `connectors/google/resolve_resource_details.go:250` — Gmail thread-fetch errors swallowed with no log. Same issue on the Microsoft side: `connectors/microsoft/resolve_resource_details.go:39` and a second silent fallback inside `connectors/microsoft/thread.go` when the `$filter` conversation query fails.
- `connectors/microsoft/thread.go:90` — `$top=100` + `receivedDateTime asc` truncates conversations to the **oldest** 100 messages, dropping the most recent replies (the ones the agent is actually replying to). Either switch to `desc` + reverse, or follow `@odata.nextLink`.

**Medium — code hygiene**
- `connectors/html_plain.go` `StripHTMLToPlain` vs `connectors/google/email_helpers.go:22-33` `htmlToPlainText` — consolidate to one shared implementation.
- `connectors/microsoft/resolve_resource_details.go:27` — `fetchGraphMessage` is called twice for the same `message_id`; pass the already-fetched `msg` into `buildMicrosoftEmailThread` or drop the outer fetch.
- `connectors/google/gmail_thread.go:42` — `sort.Slice` isn't stable; use `sort.SliceStable` for deterministic ordering when two messages share an `internalDate`.
- `connectors/microsoft/send_email_reply.go:64` — `if ct == "Text"` compares against a magic string from `graphBodyContentType`; a typo in either spot silently sends HTML as a `message.body` payload. Use a typed constant or check `*params.HTML` directly.

**Low — docs**
- `connectors/google/read_email.go:289` — the `decodeRFC5987` docstring example got mangled in fix commit `20a6c51` (`''` → U+201D `”`). One-character fix.

### Review History
| 1 | 10 | Initial review — 1 critical, 4 high, 4 medium, 1 low |
| 2 | 1 | Fix commit `20a6c51` resolved 2 items (Critical `DisplayTemplate`, Low dead code). 8 prior findings still open; flagged 1 new docstring regression. |

### Overall Assessment
Solid Phase 1 foundation. The `EmailThreadPayload` abstraction is well-designed — the shared schema + per-provider builders + 20 KiB truncation cap cleanly separates concerns and the test coverage on the truncation/multipart happy paths is good.

The biggest merge-time concern is the **pagination bug in `buildMicrosoftEmailThread`** (oldest-100 ordering): for any conversation longer than 100 messages, the approval UI will show the wrong messages — not the ones the agent is replying to. That's worth a prompt follow-up. The observability gaps are also worth addressing because a missing `email_thread` in the UI currently has no audit trail, which will make production debugging slow.

Everything else is code-hygiene cleanup. None of it is load-bearing, but the `sort.Slice` stability fix and the HTML-stripper DRY consolidation are quick wins.

Nice work getting the critical CI-breaker cleaned up quickly and throwing in the base64/attachment-filename improvements while you were in there.

---

---

Issue #981, opened 2026-04-21
Show Slack channel/thread/DM context on approval screen for Slack actions (web + mobile)


chiedoclaude:
## Problem

When a Slack-connector action reaches the approval screen (web and mobile), the reviewer sees only the outgoing payload — e.g. the message text the agent wants to post, or the channel it wants to archive. There's no visibility into **what's actually happening in that channel/thread/DM right now**, so the reviewer has to leave the app, open Slack, find the channel, and scroll history before they can approve with confidence.

Unlike email — where there's always a structural `thread_id`/`message_id` anchor (#975) — Slack has three regimes:

- **Thread replies** (`thread_ts` set) — unambiguous anchor, fetch via `conversations.replies`.
- **Top-level channel messages** — no structural anchor; often a social response to recent activity.
- **DMs** — a running 1:1 conversation with no thread concept.

We should surface the best available Slack context — thread, recent channel activity, or DM history — directly on the approval screen, normalized into one payload shape that covers all Slack actions that mutate or reference channel content.

## Scope

Applies to **all 12 Slack actions** that currently produce approval requests and benefit from surrounding context:

- **Message lifecycle (5):** `slack.send_message`, `slack.send_dm`, `slack.schedule_message`, `slack.update_message`, `slack.delete_message`
- **Secondary (7):** `slack.add_reaction`, `slack.remove_reaction`, `slack.pin_message`, `slack.unpin_message`, `slack.archive_channel`, `slack.invite_to_channel`, `slack.remove_from_channel`

Applies to **both** the web approval dialog (`ReviewApprovalDialog.tsx`) and the mobile approval detail screen (`ApprovalDetailScreen.tsx`).

## Design decisions (already resolved)

- **Tiered context strategy.** For each action, the backend picks the best available scope:
  - If `thread_ts` present → full thread (cap 20 messages).
  - Else if agent supplies an optional `in_response_to_ts` hint → anchor recent context around that message.
  - Else → recent channel window (cap 10 messages AND last 24h, whichever is smaller).
  - DMs → last 10 messages AND last 24h.
  - Always include channel metadata and a Slack deep link so the reviewer can jump in and investigate directly.
- **Privacy scope:** always show full context. No org-level toggle, no channel-membership check on the approver. Documented as a deliberate choice — the agent has the token, so the data is already accessible via the agent path.
- **Body truncation:** 4 KB per message with a `truncated: true` flag.
[...]

PR #986, opened 2026-04-21
Slack approval context: OpenAPI schema + slack/context helpers (Phase 1 / #981)

chiedocursor:
## Summary

Implements **PR 1 / Phase 1** from #981: lock the `slack_context` shape in OpenAPI, regenerate the bundled spec, and add `connectors/slack/context` helpers (with unit tests using a mock Slack API) for use in follow-up PRs.

Also wires optional **`thread_ts`** and **`in_response_to_ts`** on `slack.send_message` and `slack.schedule_message` (manifest + execution). `in_response_to_ts` is validated and stored in parameters only — it is **not** sent to Slack; PR2 will use it to anchor context. `thread_ts` is forwarded to `chat.postMessage` / `chat.scheduleMessage` when set.

`SlackConnector` exposes **`Post`** and **`Get`** so the context package can reuse the same HTTP/auth/rate-limit behavior as actions.

## OpenAPI

- New `spec/openapi/components/schemas/slack_context.yaml` and `details.slack_context` under `ActionContext.details` in `shared.yaml`.
- Root `openapi.yaml` registers the new schemas; `openapi.bundle.yaml` updated via Redocly bundle.

Regenerate ignored client types locally: `cd frontend && npm run generate:api` (and `mobile` if needed).

## Test plan

- [x] `cd frontend && npm test` (1020 tests) after regenerating API types from the bundled spec.
- [ ] `go test ./connectors/slack/context/...` — **not run in this environment** (Go toolchain mismatch vs `go.mod`); CI should run full backend tests.

## Part of

#981 — Phase 1 (Foundation) only; does not complete the full issue.

<div><a href="https://cursor.com/agents/bc-3e833edd-75da-4c71-8771-6b0509dacf8d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3e833edd-75da-4c71-8771-6b0509dacf8d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/connectors/slack/context/fetch_helpers.go
+++ b/connectors/slack/context/fetch_helpers.go
@@ -1,0 +1,226 @@
+package context
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"time"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+)
+
+const (
+	recentWindowHours = 24
+	recentCapChannel  = 10
+	recentCapDM       = 10
+	threadCap         = 20
+)
+
+// RecentMessagesOpts configures fetchRecentMessages.
+type RecentMessagesOpts struct {
+	// AnchorTS, when set, biases the window around this message (in_response_to_ts).
+	AnchorTS string
+}
+
+func messageToContextMessage(ctx context.Context, api SlackAPI, creds connectors.Credentials, teamDomain, channelID string, msg slackMessage, mcache *MentionCache) (ContextMessage, error) {
+	text := msg.Text
+	resolved, err := ResolveMentions(ctx, text, api, creds, mcache)
+	if err != nil {
+		return ContextMessage{}, err
+	}
+	body, trunc := TruncateBody(resolved)
+	var uref *UserRef
+	if msg.User != "" {
+		uref, err = fetchUserRef(ctx, api, creds, msg.User)
+		if err != nil {
+			return ContextMessage{}, err
+		}
+	}
+	isBot := msg.BotID != ""
+	cm := ContextMessage{
+		User:      uref,
+		Text:      body,
+		TS:        msg.TS,
+		Permalink: BuildPermalink(teamDomain, channelID, msg.TS),
+		IsBot:     isBot,
+		Truncated: trunc,
+		Files:     ExtractFiles(msg),
+	}
+	return cm, nil
+}
+
+func sortMessagesOldestFirst(msgs []slackMessage) {
+	sort.SliceStable(msgs, func(i, j int) bool {
+		si, ni, okI := parseSlackTSNs(msgs[i].TS)
+		sj, nj, okJ := parseSlackTSNs(msgs[j].TS)
+		if !okI || !okJ {
+			return msgs[i].TS < msgs[j].TS
+		}
+		if si != sj {
+			return si < sj
+		}
+		return ni < nj
+	})
+}
+
+// FetchRecentMessages returns up to 10 messages from the last 24h in a channel,
+// oldest first. Optional AnchorTS narrows the window around that message.
+func FetchRecentMessages(ctx context.Context, api SlackAPI, channelID string, creds connectors.Credentials, cache *sessionCache, opts RecentMessagesOpts, mcache *MentionCache) ([]ContextMessage, error) {
+	if err := validateChannelID(channelID); err != nil {
+		return nil, err
+	}
+	sess, err := resolveSession(ctx, api, creds, cache)
+	if err != nil {
+		return nil, err
+	}
+	cutoff := time.Now().Add(-recentWindowHours * time.Hour).UTC()
+	oldestTS := fmt.Sprintf("%d.%06d", cutoff.Unix(), cutoff.Nanosecond()/1000)
+
+	body := readChannelHistoryRequest{
+		Channel: channelID,
+		Limit:   100,
+		Oldest:  oldestTS,
+	}
+	var resp messagesResponse
+	if err := api.Post(ctx, "conversations.history", creds, body, &resp); err != nil {
+		return nil, err
+	}
+	if !resp.OK {
+		return nil, mapSlackErr(resp.Error)
+	}
+	raw := filterMessagesByAge(resp.Messages, cutoff)
+	sortMessagesOldestFirst(raw)
+	raw = sliceAroundAnchor(raw, opts.AnchorTS, recentCapChannel)
+	if len(raw) > recentCapChannel {
+		raw = raw[len(raw)-recentCapChannel:]
+	}
+	out := make([]ContextMessage, 0, len(raw))
+	for _, m := range raw {
+		cm, err := messageToContextMessage(ctx, api, creds, sess.TeamDomain, channelID, m, mcache)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, cm)
+	}
+	return out, nil
+}
+
+func filterMessagesByAge(msgs []slackMessage, cutoff time.Time) []slackMessage {
+	var out []slackMessage
+	for _, m := range msgs {
+		sec, nsec, ok := parseSlackTSNs(m.TS)
+		if !ok {
+			continue
+		}
+		t := time.Unix(sec, nsec).UTC()
+		if !t.Before(cutoff) {
+			out = append(out, m)
+		}
+	}
+	return out
+}
+
+// FetchThread returns the parent message and up to 19 most recent replies (20 total),
+// oldest first for the reply list (parent separate).
+func FetchThread(ctx context.Context, api SlackAPI, channelID, threadTS string, creds connectors.Credentials, cache *sessionCache, mcache *MentionCache) (parent ContextMessage, replies []ContextMessage, truncated bool, err error) {
+	if err := validateChannelID(channelID); err != nil {
+		return ContextMessage{}, nil, false, err
+	}
+	if threadTS == "" {
+		return ContextMessage{}, nil, false, &connectors.ValidationError{Message: "missing thread_ts for Slack thread context"}
+	}
+	sess, err := resolveSession(ctx, api, creds, cache)
+	if err != nil {
+		return ContextMessage{}, nil, false, err
+	}
+	body := readThreadRequest{
+		Channel: channelID,
+		TS:      threadTS,
+		Limit:   1000,
+	}
+	var resp messagesResponse
+	if err := api.Post(ctx, "conversations.replies", creds, body, &resp); err != nil {
+		return ContextMessage{}, nil, false, err
+	}
+	if !resp.OK {
+		return ContextMessage{}, nil, false, mapSlackErr(resp.Error)
+	}
+	if len(resp.Messages) == 0 {
+		return ContextMessage{}, nil, false, &connectors.ExternalError{StatusCode: 200, Message: "Slack thread has no messages"}
+	}
+	sortMessagesOldestFirst(resp.Messages)
+	parentSlack := resp.Messages[0]
+	pmsg, err := messageToContextMessage(ctx, api, creds, sess.TeamDomain, channelID, parentSlack, mcache)
+	if err != nil {
+		return ContextMessage{}, nil, false, err
+	}
+	replySlack := resp.Messages[1:]
+	if len(replySlack) > threadCap-1 {
+		truncated = true
+		replySlack = replySlack[len(replySlack)-(threadCap-1):]
+	}
+	replies = make([]ContextMessage, 0, len(replySlack))
+	for _, m := range replySlack {
+		cm, err := messageToContextMessage(ctx, api, creds, sess.TeamDomain, channelID, m, mcache)
+		if err != nil {
+			return ContextMessage{}, nil, false, err
+		}
+		replies = append(replies, cm)
+	}
+	return pmsg, replies, truncated, nil
+}
+
+// FetchDMHistory loads DM history for an IM channel opened with the given peer user ID.
+// It returns SentinelSelfDM or SentinelFirstContact when those cases apply (no messages).
+func FetchDMHistory(ctx context.Context, api SlackAPI, peerUserID string, creds connectors.Credentials, cache *sessionCache, mcache *MentionCache) (sentinel DMHistorySentinel, messages []ContextMessage, imChannelID string, err error) {
+	if peerUserID == "" {
+		return SentinelNone, nil, "", &connectors.ValidationError{Message: "missing user id for DM context"}
+	}
+	sess, err := resolveSession(ctx, api, creds, cache)
+	if err != nil {
+		return SentinelNone, nil, "", err
+	}
+	if peerUserID == sess.UserID {
+		return SentinelSelfDM, nil, "", nil
+	}
+	openBody := conversationsOpenRequest{Users: peerUserID}
+	var openResp conversationsOpenResponse
+	if err := api.Post(ctx, "conversations.open", creds, openBody, &openResp); err != nil {
+		return SentinelNone, nil, "", err
+	}
+	if !openResp.OK {
+		return SentinelNone, nil, "", mapSlackErr(openResp.Error)
+	}
+	imChannelID = openResp.Channel.ID
+
+	cutoff := time.Now().Add(-recentWindowHours * time.Hour).UTC()
+	oldestTS := fmt.Sprintf("%d.%06d", cutoff.Unix(), cutoff.Nanosecond()/1000)
+	histBody := readChannelHistoryRequest{
+		Channel: imChannelID,
+		Limit:   recentCapDM + 1,
+		Oldest:  oldestTS,
+	}
+	var histResp messagesResponse
+	if err := api.Post(ctx, "conversations.history", creds, histBody, &histResp); err != nil {
+		return SentinelNone, nil, imChannelID, err
+	}
+	if !histResp.OK {
+		return SentinelNone, nil, imChannelID, mapSlackErr(histResp.Error)
+	}
+	raw := filterMessagesByAge(histResp.Messages, cutoff)
+	sortMessagesOldestFirst(raw)
+	if len(raw) == 0 {
+		return SentinelFirstContact, nil, imChannelID, nil
+	}
+	if len(raw) > recentCapDM {
+		raw = raw[len(raw)-recentCapDM:]
+	}
+	for _, m := range raw {
+		cm, err := messageToContextMessage(ctx, api, creds, sess.TeamDomain, imChannelID, m, mcache)
+		if err != nil {
+			return SentinelNone, nil, imChannelID, err
+		}
+		messages = append(messages, cm)
+	}
+	return SentinelNone, messages, imChannelID, nil
+}
--- a/connectors/slack/context/mentions.go
+++ b/connectors/slack/context/mentions.go
@@ -1,0 +1,148 @@
+package context
+
+import (
+	"context"
+	"regexp"
+	"strings"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+)
+
+// MentionCache stores resolved display strings for a single approval request.
+type MentionCache struct {
+	Users    map[string]string // Slack user ID -> @handle
+	Channels map[string]string // Slack channel ID -> #name
+}
+
+func (m *MentionCache) userKey(id string) string {
+	if m.Users == nil {
+		m.Users = make(map[string]string)
+	}
+	return id
+}
+
+func (m *MentionCache) channelKey(id string) string {
+	if m.Channels == nil {
+		m.Channels = make(map[string]string)
+	}
+	return id
+}
+
+var (
+	reUserMention    = regexp.MustCompile(`<@([UW][A-Z0-9]+)(?:\|([^>]*))?>`)
+	reChannelMention = regexp.MustCompile(`<#(C[A-Z0-9]+)(?:\|([^>]*))?>`)
+	reSubteamLabeled = regexp.MustCompile(`<!subteam\^(S[A-Z0-9]+)\|([^>]+)>`)
+	reSubteamBare    = regexp.MustCompile(`<!subteam\^(S[A-Z0-9]+)>`)
+)
+
+// ResolveMentions replaces Slack mrkdwn mention tokens with human-readable
+// text: users and channels via API (batched with MentionCache), broadcast
+// keywords and labeled subteams without extra calls.
+func ResolveMentions(ctx context.Context, text string, api SlackAPI, creds connectors.Credentials, cache *MentionCache) (string, error) {
+	if text == "" {
+		return "", nil
+	}
+	out := text
+	out = strings.ReplaceAll(out, "<!here>", "@here")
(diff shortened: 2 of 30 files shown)


connectors/slack/context/fetch_helpers.go
```diff
@@ -0,0 +1,226 @@
...
+	if err := api.Post(ctx, "conversations.history", creds, histBody, &histResp); err != nil {
+		return SentinelNone, nil, imChannelID, err
+	}
+	if !histResp.OK {
+		return SentinelNone, nil, imChannelID, mapSlackErr(histResp.Error)
+	}
+	raw := filterMessagesByAge(histResp.Messages, cutoff)
+	sortMessagesOldestFirst(raw)
+	if len(raw) == 0 {
+		return SentinelFirstContact, nil, imChannelID, nil
```
chiedoclaude, line 213: **[High · Correctness]** `SentinelFirstContact` is returned whenever the 24h window (`cutoff = now - 24h`) is empty, not when the DM history is truly empty. An approver reviewing a DM to a peer they last talked to 30 days ago will be told this is "first contact."

**Why:** "First contact" is a material signal for approvers — it affects how the approval UI warns the user ("this is the first time you're talking to this peer, verify it's who you think"). A quiet 24h window is very different from a brand-new conversation.

**Fix:** When the 24h window comes back empty, re-probe without the `oldest` bound (`conversations.history` with `limit: 1` and no `oldest`) before returning `SentinelFirstContact`. Return `SentinelFirstContact` only when that probe also returns zero messages; otherwise return `SentinelNone` with empty `messages` so the approver UI can show "no recent activity" rather than "first contact."

---

connectors/slack/context/rate_limit.go
```diff
@@ -0,0 +1,37 @@
...
+
+// HandleRateLimit returns a metadata-only Slack context when err is or wraps a
+// Slack rate limit (HTTP 429 / RateLimitError / ratelimited). For other errors
+// it returns nil, false.
+func HandleRateLimit(err error) (*SlackContext, bool) {
+	if err == nil {
+		return nil, false
+	}
+	if !connectors.IsRateLimitError(err) {
+		return nil, false
```
chiedoclaude, line 17: **[High · Consistency]** `HandleRateLimit` only triggers on `connectors.IsRateLimitError(err)`, but `IsRateLimit` below additionally degrades on an `*ExternalError{StatusCode: 429}`. A raw 429 wrapped as `ExternalError` will be treated as rate-limited by `IsRateLimit(...)` callers but skip the metadata-only degrade in `HandleRateLimit`.

**Why:** These two helpers sit side-by-side for interchangeable use. An inconsistency here means a 429 that surfaces via `ExternalError` will show a full error to the approver instead of the documented metadata-only fallback.

**Fix:** Share one predicate — the simplest change:
```go
func HandleRateLimit(err error) (*SlackContext, bool) {
    if !IsRateLimit(err) {
        return nil, false
    }
    return &SlackContext{ContextScope: ScopeMetadataOnly}, true
}
```
Or push the 429-on-`ExternalError` check into `connectors.IsRateLimitError` so both helpers see the same definition.

---

connectors/slack/context/slack_err.go
```diff
@@ -0,0 +1,20 @@
...
+func mapSlackErr(code string) error {
+	switch code {
+	case "not_authed", "invalid_auth", "token_revoked", "token_expired", "account_inactive":
+		return &connectors.AuthError{Message: fmt.Sprintf("Slack auth error: %s", code)}
+	case "missing_scope":
+		return &connectors.AuthError{Message: "Slack token is missing a required OAuth scope — re-authorize the Slack connection"}
+	case "ratelimited":
+		return &connectors.RateLimitError{Message: "Slack API rate limit exceeded"}
+	default:
+		return &connectors.ExternalError{StatusCode: 200, Message: fmt.Sprintf("Slack API error: %s", code)}
```
chiedoclaude, line 18: **[High · Correctness]** The default branch returns `&connectors.ExternalError{StatusCode: 200, ...}` for any unknown Slack error code. `StatusCode: 200` propagates through to callers that reasonably use it for retry / audit decisions — e.g. `errors.As(err, &ext); ext.StatusCode >= 500` will never match genuine 5xx-class Slack failures surfaced this way.

**Why:** The production `mapSlackError` in `connectors/slack/slack.go` has the same pattern — already latent — and this helper now propagates it. A Slack response like `ok: false, error: "internal_error"` is categorized as `StatusCode: 200`, masking real infrastructure problems.

**Fix:** Drop the hard-coded `StatusCode: 200` in these constructors (it is not the HTTP status). Either omit the field so it defaults to `0`, or thread through the actual `resp.StatusCode` where callers have it. Apply the same fix in `connectors/slack/slack.go: mapSlackError` default case to keep parity.

---

connectors/slack/context/fetch_helpers.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, err
+	}
+	if !resp.OK {
+		return nil, mapSlackErr(resp.Error)
+	}
+	raw := filterMessagesByAge(resp.Messages, cutoff)
+	sortMessagesOldestFirst(raw)
+	raw = sliceAroundAnchor(raw, opts.AnchorTS, recentCapChannel)
+	if len(raw) > recentCapChannel {
+		raw = raw[len(raw)-recentCapChannel:]
```
chiedoclaude, line 95: **[Medium · Maintainability]** The cap `if len(raw) > recentCapChannel { raw = raw[len(raw)-recentCapChannel:] }` runs after `sliceAroundAnchor` has already trimmed the window. For a non-empty `AnchorTS`, this branch is dead; for an empty `AnchorTS`, `sliceAroundAnchor` returns `msgs` unchanged and this second cap is load-bearing.

**Why:** Two capping paths hidden in one sequence are easy to misread and easy to break when limits change. The tail-take (`raw[len(raw)-recentCapChannel:]`) silently discards anchored context if `sliceAroundAnchor`'s contract ever changes (e.g. returning a larger centered window).

**Fix:** Make `sliceAroundAnchor` the single authoritative capper by applying the cap unconditionally (even when `AnchorTS` is empty), so the caller doesn't need a trailing `if len(raw) > …`. Or invert: tail-cap first in the empty-anchor case, then call `sliceAroundAnchor` only when anchor is set. Either way, eliminate the redundant cap.

---

connectors/slack/context/mentions.go
```diff
@@ -0,0 +1,148 @@
...
+	Users    map[string]string // Slack user ID -> @handle
+	Channels map[string]string // Slack channel ID -> #name
+}
+
+func (m *MentionCache) userKey(id string) string {
+	if m.Users == nil {
+		m.Users = make(map[string]string)
+	}
+	return id
+}
```
chiedoclaude, line 22: **[Medium · Maintainability]** `userKey`/`channelKey` look like pure key-derivation helpers but silently mutate the receiver by lazy-initializing `m.Users`/`m.Channels`. The callsite `cache.Users[cache.userKey(uid)] = label` only works because of that hidden side effect — a reader can't tell from the name.

**Why:** This is exactly the shape that bites the next maintainer when they "simplify" one callsite and drop the side effect, producing a nil-map panic. There's also no mutex, so calling `ResolveMentions` concurrently with a shared cache is a data race on the map (Go's race detector will flag it).

**Fix:** Remove the hidden init. Give `MentionCache` a small constructor and initialize eagerly:
```go
func NewMentionCache() *MentionCache {
    return &MentionCache{Users: map[string]string{}, Channels: map[string]string{}}
}
```
Delete `userKey`/`channelKey`. If callers can share a cache across goroutines (e.g. parallel RPCs within a single approval flow), add a `sync.Mutex` — even a simple one avoids race-detector alarms in tests.

---

connectors/slack/context/session.go
```diff
@@ -0,0 +1,74 @@
...
+	return info, nil
+}
+
+func teamDomainFromAuthURL(raw string) (string, error) {
+	u, err := url.Parse(strings.TrimSpace(raw))
+	if err != nil || u.Host == "" {
+		return "", &connectors.ExternalError{StatusCode: 200, Message: "Slack auth.test url could not be parsed"}
+	}
+	host := u.Hostname()
+	host = strings.TrimSuffix(host, ".slack.com")
```
chiedoclaude, line 69: **[Medium · Correctness]** `teamDomainFromAuthURL` assumes URLs of the form `https://<team>.slack.com/`. Enterprise Grid workspaces often return `https://<team>.enterprise.slack.com/`; `TrimSuffix(host, ".slack.com")` leaves `<team>.enterprise` — which then produces broken `BuildPermalink`/`BuildChannelPermalink` values since Slack archive URLs don't use that host form.

**Why:** Any Enterprise Grid customer ends up with non-clickable `permalink` values in the approval context — one of the core value props of this payload.

**Fix:** Take only the leftmost label as the team domain, independent of grid vs. non-grid:
```go
host := u.Hostname()
if i := strings.Index(host, "."); i > 0 {
    host = host[:i]
}
```
And add a test covering `https://acme.enterprise.slack.com/`.

---

connectors/slack/context/fetch_helpers.go
```diff
@@ -0,0 +1,226 @@
...
+func messageToContextMessage(ctx context.Context, api SlackAPI, creds connectors.Credentials, teamDomain, channelID string, msg slackMessage, mcache *MentionCache) (ContextMessage, error) {
+	text := msg.Text
+	resolved, err := ResolveMentions(ctx, text, api, creds, mcache)
+	if err != nil {
+		return ContextMessage{}, err
+	}
+	body, trunc := TruncateBody(resolved)
+	var uref *UserRef
+	if msg.User != "" {
+		uref, err = fetchUserRef(ctx, api, creds, msg.User)
```
chiedoclaude, line 34: **[Medium · Performance]** `messageToContextMessage` calls `fetchUserRef` once per message author and `ResolveMentions` calls `fetchUserRef` once per unique mentioned user. Within a single message these are serial; across messages only repeated IDs hit the cache. A thread with 20 replies plus a few @-mentions each can fan out to 50+ serial `users.info` round-trips, which will dominate approval-context latency.

**Why:** Approval context is on the critical path of human approval latency. Each `users.info` round-trip is ~150–400ms on standard tier; 50× that is very noticeable to the approver.

**Fix (follow-up is fine):** Share a single user cache between `messageToContextMessage` and `ResolveMentions` so author and mention lookups coalesce. Longer-term, collect the full set of distinct user IDs across the whole batch first, then issue lookups with bounded concurrency (e.g. `errgroup.SetLimit(4)`). Not blocking for this phase, but worth a tracked item on #981.

---

connectors/slack/schedule_message.go
```diff
@@ -141,15 +155,29 @@ func (a *scheduleMessageAction) Execute(ctx context.Context, req connectors.Acti
...
+		postErr = a.conn.doPost(ctx, "chat.scheduleMessage", req.Credentials, struct {
+			Channel  string `json:"channel"`
+			Text     string `json:"text"`
+			PostAt   int64  `json:"post_at"`
+			ThreadTS string `json:"thread_ts,omitempty"`
+		}{
+			Channel:  params.Channel,
+			Text:     params.Message,
+			PostAt:   postAtUnix,
+			ThreadTS: params.ThreadTS,
```
chiedoclaude, line 170: **[Low · Maintainability]** The `if params.ThreadTS != ""` branch duplicates the request body as an inline anonymous struct with an extra `ThreadTS` field; `send_message.go` has the same shape. As more optional fields land in phase 2, these branches multiply combinatorially.

**Why:** Adding one more optional field doubles the branches. Easy for the two copies to drift (e.g. a future `unfurl_links: false` applied in one branch but not the other).

**Fix:** Promote the request struct to a single named type with `omitempty` on optional fields and populate it once:
```go
type scheduleMessageRequest struct {
    Channel  string `json:"channel"`
    Text     string `json:"text"`
    PostAt   int64  `json:"post_at"`
    ThreadTS string `json:"thread_ts,omitempty"`
}
```
Then the `Execute` body becomes a single `doPost` call. Same treatment for `send_message.go`.

---

chiedocursor:
## 🤖 Watch Session Summary

### Changes Made
- **`Sync PR branch with origin/main`** (`7251c20d`) — watch-poll merged main into the branch (no conflicts); pushed 1d890e78 and 7251c20d to keep PR current.

- **`spec/openapi/components/schemas/shared.yaml, spec/openapi/openapi.yaml, spec/openapi/openapi.bundle.yaml`** — Merged origin/main email thread ActionContext.details (email_thread + EmailThread* schemas) with branch slack_context; both properties now documented; rebundled OpenAPI. (`9b96c8fd`)

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

## 🔧 CI / audit remediation

### Remediation round 1

- **ci.yml** — conclusion `failure` — Backend tests failed: slack/context tests used a custom mock for connectors.Credentials but Credentials is a concrete struct. Replaced with connectors.NewCredentials(map) in mock_api_test.go and call sites. (`248bd2f6`)

connectors/slack/context/permalink.go
```diff
@@ -0,0 +1,43 @@
...
+	domain = strings.TrimSuffix(domain, ".slack.com")
+	domain = strings.TrimSpace(domain)
+	if domain == "" || channelID == "" || ts == "" {
+		return ""
+	}
+	p := permalinkMessagePath(channelID, ts)
+	if p == "" {
+		return ""
+	}
+	return fmt.Sprintf("https://%s.slack.com/archives/%s/%s", domain, channelID, p)
```
chiedoclaude, line 22: **[Low · Correctness]** `BuildPermalink` / `BuildChannelPermalink` trim `.slack.com` case-sensitively. `teamDomainFromAuthURL` normalizes via `url.Hostname()`, but callers that hand-pass `teamDomain` (tests, future code paths) can feed mixed case and produce a broken permalink (or one that doesn't match Slack's lowercase-domain redirect).

**Why:** Defensive, cheap, and a single `strings.ToLower` avoids confusing test failures and subtle prod bugs.

**Fix:**
```go
domain := strings.ToLower(strings.TrimSpace(teamDomain))
domain = strings.TrimSuffix(domain, ".slack.com")
```
Apply to both `BuildPermalink` and `BuildChannelPermalink`.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 10 (1 critical, 3 high, 4 medium, 2 low)

### [Critical · Architecture] Accidental removal of EmailThread schemas from the spec

This PR removes the `email_thread` property from `ActionContext.details.properties` **and** deletes the `EmailThread`, `EmailThreadMessage`, and `EmailThreadAttachment` schemas from `spec/openapi/components/schemas/shared.yaml`. The matching `components/schemas` entries in `spec/openapi/openapi.yaml` are also gone, and `openapi.bundle.yaml` no longer carries those type definitions. The stated scope of the PR is *adding* Slack context — the email removals look unintentional.

**Why it matters:** the Go backend still populates `details.email_thread` (`connectors/email_thread.go`, `connectors/google/gmail_thread.go`, `connectors/microsoft/thread.go`, `api/approval_email_thread.go`). API consumers that relied on the documented `EmailThread*` types will see an undocumented field, and any downstream code generation off the bundled spec will lose these types.

**Fix:** restore the three `EmailThread*` schemas in `shared.yaml`, put `email_thread` back under `ActionContext.details.properties` alongside the new `slack_context`, restore the three `$ref`s in `openapi.yaml`, and re-bundle. The `details` bag should end up with *both* keys:

```yaml
properties:
  slack_context:
    $ref: './slack_context.yaml#/SlackContext'
  email_thread:
    $ref: './shared.yaml#/EmailThread'
```

(I tried to attach this as an inline comment on the `slack_context` line of `shared.yaml`, but GitHub rejected it because the critical issue spans deleted lines outside the hunk — hoisting it here so it’s not lost.)

---

### Summary

The PR delivers a clean, well-structured foundation — types align tightly with the OpenAPI schema, tests use a clear mock HTTP layer, and the action manifest/param wiring for `thread_ts`/`in_response_to_ts` is minimal and readable. The main concern is the regression above; the other findings are smaller: the rate-limit degradation path is inconsistent between `HandleRateLimit` and `IsRateLimit`; `FetchDMHistory` treats a 24h silent window as first-contact; a couple of `MentionCache` / slicing helpers have confusing shapes; and the generic `mapSlackErr` default uses `StatusCode: 200` which can mislead callers of `ExternalError`.

### What Looks Good
- `SlackContext` schema is well-scoped — required/optional fields are sensible, the `context_scope` enum captures the important DM edge cases (`self_dm`, `first_contact_dm`, `metadata_only`), and documentation is clear.
- Surfacing the existing `SlackConnector.Post`/`Get` request lifecycle (auth, rate-limit handling, timeout) as the seam for the context package is much cleaner than reimplementing HTTP in the subpackage.
- `sliceAroundAnchor` centers around the anchor correctly when the anchor falls inside the window; the algorithm is easy to verify and well-tested.
- Tests are focused and readable, and the mock API in `mock_api_test.go` is a minimal, reusable seam.
- Manifest changes are surgical — new params are hidden in the UI, validated, and still covered by the stored-parameter hash.

---

---

chiedoclaude (commented):
## Review Round 2/6

**Findings:** 0 new (Round 1: 1 critical, 3 high, 4 medium, 2 low)

### Changes Since Last Round
New commits on this PR:
- `9b96c8f` — **Merge origin/main: combine email_thread and slack_context OpenAPI details**
- `1d890e7` / `7251c20` — upstream merges from `main`
- `248bd2f` — `fix(slack/context): use connectors.NewCredentials in tests` (CI compile fix, not a review finding)

### Fixed ✅
- **[Critical · Correctness] EmailThread schema regression (shared.yaml / openapi.yaml)** — confirmed resolved in `9b96c8f`. `ActionContext.details.properties` now declares both `email_thread` and `slack_context`, and the `EmailThread`, `EmailThreadMessage`, `EmailThreadAttachment` schemas are restored in both `openapi.yaml` and `openapi.bundle.yaml`. Nice clean conflict resolution.

### Remaining Issues
Round 1 findings not addressed in the new commits. See original inline comments:
1. `[High · Correctness]` `SentinelFirstContact` returned for empty 24h window (`fetch_helpers.go:213`)
2. `[High · Consistency]` `HandleRateLimit` vs `IsRateLimit` divergence (`rate_limit.go:17`)
3. `[High · Correctness]` Default `mapSlackErr` returns `StatusCode: 200` for errors (`slack_err.go:18`)
4. `[Medium · Maintainability]` Redundant cap after `sliceAroundAnchor` (`fetch_helpers.go:95`)
5. `[Medium · Maintainability]` `userKey`/`channelKey` hidden init side-effect + non-thread-safe `MentionCache` (`mentions.go:22`)
6. `[Medium · Correctness]` Enterprise Grid (`*.enterprise.slack.com`) not handled (`session.go:69`)
7. `[Medium · Performance]` Redundant `fetchUserRef` calls (`fetch_helpers.go:34`)
8. `[Low · Maintainability]` Duplicated anonymous request struct (`schedule_message.go:170`, also in `send_message.go`)
9. `[Low · Correctness]` Case-sensitive `.slack.com` trim (`permalink.go:22`)

### What Looks Good
- Conflict resolution in `9b96c8f` was careful — both details properties coexist, examples updated accordingly, and the bundle was regenerated. Low-risk merge done well.
- Keeping `248bd2f` minimal and scoped purely to the test helper is the right instinct; no drive-by changes.

---

chiedoclaude:
## Code Review Complete

**PR:** #986 — Slack approval context: OpenAPI schema + slack/context helpers (Phase 1 / #981)
**Author:** @chiedocursor
**Rounds completed:** 4/6
**Reason for stopping:** PR merged during Round 4

### Outstanding Issues
Nine Round 1 findings were **not addressed before merge**. They should be carried into the Phase 2 PR or filed as follow-up issues:

**High severity (3)**
- `[High · Correctness]` `SentinelFirstContact` is returned whenever the 24h DM window is empty, not when the DM channel is genuinely empty — misleading signal to approvers. (`connectors/slack/context/fetch_helpers.go:213`)
- `[High · Consistency]` `HandleRateLimit` and `IsRateLimit` treat raw 429 `*ExternalError` differently; a raw 429 not wrapped in `RateLimitError` skips the degradation path. (`connectors/slack/context/rate_limit.go:17`)
- `[High · Correctness]` `mapSlackErr` default branch returns `&ExternalError{StatusCode: 200, …}` for unknown Slack error codes — callers that switch on `StatusCode` will misclassify errors as success-adjacent. (`connectors/slack/context/slack_err.go:18`)

**Medium severity (4)**
- Redundant cap after `sliceAroundAnchor` (`fetch_helpers.go:95`)
- `MentionCache.userKey`/`channelKey` hidden lazy-init side-effect + no mutex for concurrent use (`mentions.go:22`)
- `teamDomainFromAuthURL` doesn't handle Enterprise Grid `*.enterprise.slack.com` hosts (`session.go:69`)
- Redundant `fetchUserRef` calls per message — cacheable at the author level (`fetch_helpers.go:34`)

**Low severity (2)**
- Duplicated anonymous request struct for `thread_ts` in both `schedule_message.go:170` and `send_message.go` — extract
- `BuildPermalink` / `BuildChannelPermalink` trim `.slack.com` case-sensitively; callers that pass raw hostnames with mixed case get a broken URL (`permalink.go:22`)

**Recommendation:** Open a follow-up issue bundling items 1–3 (High) for the Phase 2 PR, and address 4–9 opportunistically when adjacent code is next touched.

### Review History
| 1 | 9 + 1 critical in body | Initial review — Critical `EmailThread` schema regression flagged in body (couldn't be inline because the affected lines were deletions outside the hunk); 3 High, 4 Medium, 2 Low inline |
| 2 | 0 | Summary-only review — Critical fix in `9b96c8f` confirmed; 9 prior findings unchanged |
| 3 | 0 | Clean — no new commits, no new findings |
| 4 | 0 | PR merged during wait — stopped early |

### Overall Assessment
The Critical schema regression was caught and fixed mid-review (`9b96c8f` cleanly merged both `email_thread` and `slack_context` under `ActionContext.details`), which was the most important outcome of this review — merging as-is would have broken the email-reply connectors (`connectors/email_thread.go`, Gmail, Microsoft) that depend on `EmailThread`.

The PR itself is well-structured: new functionality is cleanly separated into the `connectors/slack/context/` subpackage, OpenAPI additions are comprehensive, and the `ScopeMetadataOnly` degradation pattern is a thoughtful approach to Slack rate limits. The remaining 9 findings are real but non-blocking — most are correctness edge cases (Enterprise Grid hosts, empty DM window semantics, 429 handling divergence) that won't affect the common path but will matter for production robustness. None are so severe that blocking the merge would have been justified once the Critical was resolved.

Good work on the careful merge conflict resolution in `9b96c8f` and the minimal scope of the CI fix in `248bd2f`.

---

PR #1011, opened 2026-04-21
Slack: list_unread and mark_read for agent unread state (#1010)

### LINKED ISSUE #1010: Slack: expose unread-message state to agents
## Background

The Slack connector uses a user token (`xoxp-`) and already holds the scopes needed to read a user's unread state (`channels:read`, `groups:read`, `im:read`, `mpim:read`, plus the matching `*:history` and `*:write` scopes). However, no action exposes unread state to agents, so when a user asks "do I have unread Slack messages?" the agent can only describe recent history — it has no way to tell what the user has or hasn't read.

Slack's per-user read cursor (`last_read`) and unread counts (`unread_count_display`) are returned by `conversations.info` when called with a user token. The existing `slack.read_channel_messages` action already accepts an `oldest` parameter, so once `last_read` is available the agent can fetch the actual unread messages without any new history action.

## Scope

Add two new actions to the Slack connector. **No OAuth scope changes required** — all scopes are already requested in `connectors/slack/slack.go:41-65`.

### Phase 1 — single PR (all tasks cohesive, share a `conversations.info` helper)

- [ ] **Add `slack.list_unread` action**
  - Enumerate the user's conversations via `users.conversations` (types: `public_channel,private_channel,mpim,im`)
  - For each, call `conversations.info` and read `last_read` + `unread_count_display` + `latest`
  - Filter to channels where `unread_count_display > 0`
  - Return a list of `{channel_id, channel_name, channel_type, unread_count, last_read_ts, latest_message_preview}` where `latest_message_preview` is a short snippet of the `latest` message (`text` truncated to a reasonable length, plus `user` and `ts`). The preview lets the agent summarize "3 unread in #foo about bar" without a follow-up `conversations.history` call.
  - Register the action in `connectors/slack/register.go` and `manifest.go` (append at the end — see CLAUDE.md on minimizing merge conflicts)
  - Unit tests covering: no unreads, mixed channel types, rate-limit handling, empty channel list, missing/empty `latest` field

- [ ] **Add `slack.mark_read` action**
  - Wrap `conversations.mark` (requires only `channels:write` / `groups:write` / `im:write` / `mpim:write`, all already granted)
  - Params: `channel_id` (required), `ts` (**required** — no default-to-latest behavior, so the agent can't accidentally clear unreads it never surfaced to the user)
  - Register in `register.go` and `manifest.go`
  - Unit tests covering: successful mark, missing `ts` is rejected with a `ValidationError`, invalid channel, rate-limit handling

- [ ] **Agent-facing documentation**
  - Update `connectors/slack/README.md` with a short usage note: enumerate unreads via `slack.list_unread`, fetch messages via existing `slack.read_channel_messages` with `oldest = last_read_ts`, optionally clear with `slack.mark_read` passing the explicit `ts` of the last surfaced message
  - Make sure the action descriptions in `manifest.go` steer the agent toward this flow so it doesn't fall back to blindly calling `conversations.history`

## Non-goals

- No bot-token support (Slack intentionally walls bots off from user read state)
- No new OAuth scopes
- No new "fetch unread messages" action — the existing `slack.read_channel_messages` with `oldest=last_read_ts` covers this

## Test plan

- [ ] `make test-backend` passes (note: local sandbox can't run Go tests per CLAUDE.md — verify in CI)
- [ ] Manual check via the CLI: "do I have unread Slack messages?" returns the actual unread list, not just recent history
- [ ] Manual check: marking a channel read via `slack.mark_read` (with explicit `ts`) causes the next `slack.list_unread` to drop that channel
- [ ] Manual check: calling `slack.mark_read` without `ts` returns a validation error

chiedocursor:
## Summary

Implements GitHub issue #1010: expose Slack per-user unread state to agents without new OAuth scopes.

- **`slack.list_unread`** — Paginates `users.conversations` (types `public_channel,private_channel,mpim,im`), calls `conversations.info` per channel via shared `fetchConversationInfo`, keeps rows where `unread_count_display > 0`, returns `channel_id`, `channel_name`, `channel_type`, `unread_count`, `last_read_ts`, and optional `latest_message_preview` (`text` truncated to 200 runes, `user`/`ts`).
- **`slack.mark_read`** — `conversations.mark` with required `channel_id` and `ts` (validated); runs existing `verifyChannelAccess` before marking.
- **Manifest** — New actions + standing-approval templates; **`slack.read_channel_messages`** description updated to steer agents toward `list_unread` + `oldest=last_read_ts` instead of blind history.
- **README** — Short agent workflow for unreads / read / mark.
- **`ResolveResourceDetails`** — Resolves `channel_name` for `slack.mark_read` from `channel_id`.

## Test plan

- [ ] `make test-backend` / `go test ./connectors/slack/...` on CI (local agent environment could not run the full Go test suite due to toolchain constraints).
- [ ] Manual: CLI — “do I have unread Slack messages?” uses `list_unread` + `read_channel_messages` with `oldest`.
- [ ] Manual: `mark_read` with explicit `ts` then `list_unread` drops the channel; `mark_read` without `ts` returns validation error.

Closes #1010

<div><a href="https://cursor.com/agents/bc-2f1d4d30-4164-4503-ab29-968050c38f16"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2f1d4d30-4164-4503-ab29-968050c38f16"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/connectors/slack/conversation_info.go
+++ b/connectors/slack/conversation_info.go
@@ -1,0 +1,82 @@
+package slack
+
+import (
+	"context"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+)
+
+// conversationInfoChannel is the subset of conversations.channel used for
+// unread state and display (conversations.info).
+type conversationInfoChannel struct {
+	ID                 string        `json:"id"`
+	Name               string        `json:"name"`
+	User               string        `json:"user"`
+	IsPrivate          bool          `json:"is_private"`
+	IsIM               bool          `json:"is_im"`
+	IsMPIM             bool          `json:"is_mpim"`
+	LastRead           string        `json:"last_read"`
+	UnreadCountDisplay int           `json:"unread_count_display"`
+	Latest             *slackMessage `json:"latest"`
+}
+
+type conversationInfoResponse struct {
+	slackResponse
+	Channel conversationInfoChannel `json:"channel"`
+}
+
+// fetchConversationInfo loads a single channel via GET conversations.info.
+func (c *SlackConnector) fetchConversationInfo(ctx context.Context, creds connectors.Credentials, channelID string) (*conversationInfoChannel, error) {
+	var resp conversationInfoResponse
+	if err := c.doGet(ctx, "conversations.info", creds, map[string]string{"channel": channelID}, &resp); err != nil {
+		return nil, err
+	}
+	if !resp.OK {
+		return nil, resp.asError()
+	}
+	return &resp.Channel, nil
+}
+
+const maxLatestPreviewRunes = 200
+
+func channelDisplayName(ch *conversationInfoChannel) string {
+	if ch == nil {
+		return ""
+	}
+	if ch.Name != "" {
+		return ch.Name
+	}
+	if ch.IsIM && ch.User != "" {
+		return ch.User
+	}
+	return ""
+}
+
+func channelTypeLabel(ch *conversationInfoChannel) string {
+	if ch == nil {
+		return "unknown"
+	}
+	switch {
+	case ch.IsIM:
+		return "im"
+	case ch.IsMPIM:
+		return "mpim"
+	case ch.IsPrivate:
+		return "private_channel"
+	default:
+		return "public_channel"
+	}
+}
+
+func truncatePreviewText(s string) string {
+	if s == "" {
+		return ""
+	}
+	if utf8.RuneCountInString(s) <= maxLatestPreviewRunes {
+		return strings.TrimSpace(s)
+	}
+	runes := []rune(s)
+	return strings.TrimSpace(string(runes[:maxLatestPreviewRunes])) + "…"
+}
--- a/connectors/slack/list_unread.go
+++ b/connectors/slack/list_unread.go
@@ -1,0 +1,128 @@
+package slack
+
+import (
+	"context"
+
+	"github.com/supersuit-tech/permission-slip/connectors"
+)
+
+// listUnreadAction implements slack.list_unread.
+type listUnreadAction struct {
+	conn *SlackConnector
+}
+
+type listUnreadParams struct{}
+
+func (p *listUnreadParams) validate() error { return nil }
+
+type latestMessagePreview struct {
+	Text string `json:"text,omitempty"`
+	User string `json:"user,omitempty"`
+	TS   string `json:"ts,omitempty"`
+}
+
+type unreadChannelEntry struct {
+	ChannelID            string                `json:"channel_id"`
+	ChannelName          string                `json:"channel_name,omitempty"`
+	ChannelType          string                `json:"channel_type"`
+	UnreadCount          int                   `json:"unread_count"`
+	LastReadTS           string                `json:"last_read_ts,omitempty"`
+	LatestMessagePreview *latestMessagePreview `json:"latest_message_preview,omitempty"`
+}
+
+type listUnreadResult struct {
+	UnreadChannels []unreadChannelEntry `json:"unread_channels"`
+}
+
+func (a *listUnreadAction) Execute(ctx context.Context, req connectors.ActionRequest) (*connectors.ActionResult, error) {
+	var params listUnreadParams
+	if err := parseAndValidate(req.Parameters, &params); err != nil {
+		return nil, err
+	}
+
+	if req.UserEmail == "" {
+		return nil, &connectors.ValidationError{
+			Message: "listing unread conversations requires your Permission Slip profile to have an email address matching your Slack account",
+		}
+	}
+
+	slackUserID, err := a.conn.lookupSlackUserByEmail(ctx, req.Credentials, req.UserEmail)
+	if err != nil {
+		return nil, err
+	}
+	if slackUserID == "" {
+		return nil, &connectors.ValidationError{
+			Message: "no Slack user found matching your email — ensure your Permission Slip email matches your Slack account",
+		}
+	}
+
+	types := "public_channel,private_channel,mpim,im"
+	var entries []unreadChannelEntry
+	cursor := ""
+	for page := 0; page < maxUserConversationPages; page++ {
+		body := usersConversationsRequest{
+			User:   slackUserID,
+			Types:  types,
+			Limit:  200,
+			Cursor: cursor,
+		}
+		var resp usersConversationsResponse
+		if err := a.conn.doPost(ctx, "users.conversations", req.Credentials, body, &resp); err != nil {
+			return nil, err
+		}
+		if !resp.OK {
+			return nil, resp.asError()
+		}
+
+		for _, ch := range resp.Channels {
+			info, err := a.conn.fetchConversationInfo(ctx, req.Credentials, ch.ID)
+			if err != nil {
+				return nil, err
+			}
+			if info.UnreadCountDisplay <= 0 {
+				continue
+			}
+			var previewPtr *latestMessagePreview
+			if info.Latest != nil {
+				p := latestMessagePreview{
+					Text: truncatePreviewText(info.Latest.Text),
+					User: info.Latest.User,
+					TS:   info.Latest.TS,
+				}
+				if info.Latest.BotID != "" && p.User == "" {
+					p.User = info.Latest.BotID
+				}
+				previewPtr = &p
+			}
+			name := channelDisplayName(info)
+			if name == "" {
+				name = channelDisplayNameFromListEntry(ch)
+			}
+			entries = append(entries, unreadChannelEntry{
+				ChannelID:            info.ID,
+				ChannelName:          name,
+				ChannelType:          channelTypeLabel(info),
+				UnreadCount:          info.UnreadCountDisplay,
+				LastReadTS:           info.LastRead,
+				LatestMessagePreview: previewPtr,
+			})
+		}
+
+		if resp.Meta == nil || resp.Meta.NextCursor == "" {
+			break
+		}
+		cursor = resp.Meta.NextCursor
+	}
+
+	return connectors.JSONResult(listUnreadResult{UnreadChannels: entries})
+}
+
+func channelDisplayNameFromListEntry(ch listChannelEntry) string {
+	if ch.Name != "" {
+		return ch.Name
+	}
+	if ch.IsIM && ch.User != "" {
+		return ch.User
+	}
+	return ""
+}
--- a/connectors/slack/manifest.go
+++ b/connectors/slack/manifest.go
@@ -134,7 +134,7 @@ func (c *SlackConnector) Manifest() *connectors.ConnectorManifest {
 			{
 				ActionType:      "slack.read_channel_messages",
 				Name:            "Read Channel Messages",
-				Description:     "Read recent messages from a Slack channel, DM (D…), or group DM (G…).",
+				Description:     "Read recent messages from a Slack channel, DM (D…), or group DM (G…). To fetch only unread messages, call slack.list_unread first and pass oldest = last_read_ts from that result (do not rely on blind recent history).",
 				RiskLevel:       "low",
 				DisplayTemplate: "Read messages from {{channel_name}}",
 				ParametersSchema: json.RawMessage(connectors.TrimIndent(`{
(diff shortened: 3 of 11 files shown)


chiedocursor:
## 🤖 Watch Session Summary

### Changes Made

### Decision Log

#### ✅ Implemented
No review comments acted on.

---

connectors/slack/list_unread.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		var resp usersConversationsResponse
+		if err := a.conn.doPost(ctx, "users.conversations", req.Credentials, body, &resp); err != nil {
+			return nil, err
+		}
+		if !resp.OK {
+			return nil, resp.asError()
+		}
+
+		for _, ch := range resp.Channels {
```
chiedoclaude, line 77: **[High · Performance]** N+1 sequential `conversations.info` per channel is a hard scalability wall.

`users.conversations` with `types="public_channel,private_channel,mpim,im"` returns every conversation the user belongs to — in many workspaces that is hundreds of public channels (everyone is auto-added to company-wide ones). For each, we make a synchronous `conversations.info` call. Slack's tier for `conversations.info` is ~50 req/min per token, so any user in more than ~50 conversations will almost certainly trip rate-limiting mid-enumeration.

**Why it matters:** when the rate-limit hits (or any per-channel error fires — see line 79), the loop returns `nil, err` and the agent gets nothing — none of the unreads we already collected. Combined with the 30s HTTP client timeout, a moderately active user's "do I have unread Slack messages?" question will fail in most workspaces.

**Fix — one or more of:**

1. **Return partial results on transient failure.** Convert per-channel rate-limit / external errors into a `truncated: true` flag plus whatever we collected so far, so the agent can still surface something.
2. **Bounded parallelism** for `conversations.info` (e.g. 4–8 workers with `errgroup.Group.SetLimit`) to cut wall time by ~N×.
3. **Short-circuit** — add an optional `max_results` parameter so an agent asking "do I have any unreads" can stop after, say, 20 hits instead of scanning the full workspace.

At minimum, (1) is important — the current "abort the entire result set on one 429" behavior makes this action practically unusable at scale.

---

connectors/slack/list_unread.go
```diff
@@ -0,0 +1,128 @@
...
+					Text: truncatePreviewText(info.Latest.Text),
+					User: info.Latest.User,
+					TS:   info.Latest.TS,
+				}
+				if info.Latest.BotID != "" && p.User == "" {
+					p.User = info.Latest.BotID
+				}
+				previewPtr = &p
+			}
+			name := channelDisplayName(info)
```
chiedoclaude, line 97: **[High · Correctness]** Silent pagination truncation hides unreads with no caller signal.

`maxUserConversationPages = 50` × `Limit: 200` = 10k channels. When the loop hits the cap, we just `break` and return whatever was collected so far — the result payload has no `truncated` / `has_more` field, so an agent that asks "do I have any unreads" gets back `unread_channels: []` and confidently tells the user "you're all caught up" even though pages 51+ were never inspected.

Existing `ListChannels` has the same truncation behavior, but it's visible to the caller because they're asking for *the list of channels* — here the caller is asking a question whose answer is false-negative-biased by the cap.

**Fix:** thread the cursor/cap result back in the response, e.g.:

```go
type listUnreadResult struct {
    UnreadChannels []unreadChannelEntry `json:"unread_channels"`
    Truncated      bool                 `json:"truncated,omitempty"`
    NextCursor     string               `json:"next_cursor,omitempty"`
}
```

Set `Truncated = true` when we exit the loop because of `page == maxUserConversationPages-1` (not because `NextCursor == ""`), and pass `NextCursor` through so agents can paginate. Update the README + manifest description to say the result may be truncated and how to continue.

---

connectors/slack/manifest.go
```diff
@@ -1003,6 +1043,22 @@ func (c *SlackConnector) Manifest() *connectors.ConnectorManifest {
...
+				StandingApproval: neverExpire,
+			},
+			{
+				ID:               "tpl_slack_mark_read",
+				ActionType:       "slack.mark_read",
+				Name:             "Mark conversations read",
+				Description:      "Agent can mark a conversation read at an explicit message timestamp.",
+				Parameters:       json.RawMessage(`{"channel_id":"*","ts":"*"}`),
+				StandingApproval: neverExpire,
+			},
```
chiedoclaude, line 1061: **[Medium · Security]** Standing approval with wildcards on `slack.mark_read` is over-privileged.

`Parameters: {"channel_id":"*","ts":"*"}` + `StandingApproval: neverExpire` means once the user grants this template, *any* agent that holds the Slack connection can silently clear the user's read cursor in *any* conversation to *any* timestamp, forever — without ever prompting the user again.

**Why it matters:** unlike a read/list action, `mark_read` mutates human-visible state. A misbehaving or compromised agent could hide DMs the user never saw (mark an important 1:1 thread fully read so the unread badge disappears), and the safeguard cited in the description ("ts = last message you surfaced") is entirely agent-honor-system — nothing at the Permission Slip layer enforces that the agent actually surfaced those messages.

**Fix — pick one:**

1. **Drop the standing-approval template for `mark_read`.** The primary benefit of standing approval is saving user friction on repeated *idempotent reads*; `mark_read` is neither. Agents can still call it, they'll just need per-request approval.
2. **Or narrow the wildcard.** If you want to keep friction low for automation, at least require the user to enumerate channel IDs in the standing approval (drop `"channel_id":"*"`). Wildcard on `ts` alone is a much smaller blast radius than wildcard on both.

This is a trust/UX decision so flagging rather than marking critical — but "silently hide notifications from the user" is a capability worth gating more carefully than `list_channels`.

---

connectors/slack/conversation_info.go
```diff
@@ -0,0 +1,82 @@
...
+// fetchConversationInfo loads a single channel via GET conversations.info.
+func (c *SlackConnector) fetchConversationInfo(ctx context.Context, creds connectors.Credentials, channelID string) (*conversationInfoChannel, error) {
+	var resp conversationInfoResponse
+	if err := c.doGet(ctx, "conversations.info", creds, map[string]string{"channel": channelID}, &resp); err != nil {
+		return nil, err
+	}
+	if !resp.OK {
+		return nil, resp.asError()
+	}
+	return &resp.Channel, nil
```
chiedoclaude, line 39: **[Medium · Maintainability]** `fetchConversationInfo` is a good consolidation, but two existing callers still roll their own.

After this PR lands, the package has *three* independent `conversations.info` GET implementations with bespoke response structs:

- `isChannelPrivate` in `channel_membership.go` (only reads `is_private`)
- `resolveChannel` in `resolve_resource_details.go` (only reads `name`, `is_private`)
- `fetchConversationInfo` here (reads the full `conversationInfoChannel`)

The comment on this file even advertises it as the shared helper — let's actually make it shared. Both existing callers read a strict subset of `conversationInfoChannel`, so migrating them to `fetchConversationInfo` is a small, safe refactor and eliminates two parallel definitions of the same response shape.

Happy to land this as a follow-up PR if you'd rather keep this one tightly scoped — just flagging so it doesn't silently rot into three-way drift.

---

connectors/slack/list_unread.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func channelDisplayNameFromListEntry(ch listChannelEntry) string {
+	if ch.Name != "" {
+		return ch.Name
+	}
+	if ch.IsIM && ch.User != "" {
+		return ch.User
+	}
+	return ""
```
chiedoclaude, line 127: **[Low · Maintainability]** `channelDisplayNameFromListEntry` duplicates `channelDisplayName` for a sibling type.

Both functions implement the same three-line fallback (`Name` → `User` for IMs → `""`) against two structurally-identical types (`listChannelEntry` vs `*conversationInfoChannel`). Since `listChannelEntry` already has `Name`/`User`/`IsIM` fields at the same positions, you can collapse these to one helper accepting either a small interface or explicit string args:

```go
func displayNameFor(name, user string, isIM bool) string {
    if name != "" { return name }
    if isIM && user != "" { return user }
    return ""
}
```

Then both call sites read `displayNameFor(info.Name, info.User, info.IsIM)` / `displayNameFor(ch.Name, ch.User, ch.IsIM)`. Minor, not a blocker.

---

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 5 (0 critical, 2 high, 2 medium, 1 low)

The action shape and test coverage are in good shape — the implementation is idiomatic Go, uses the existing `doGet` / `parseAndValidate` / `verifyChannelAccess` plumbing correctly, and the tests cover the key branches (no-email, no-unreads, mixed types + rune-level preview truncation, rate-limit surface, pagination). The core concerns are about **how this behaves at workspace scale** and one **security/trust trade-off** on the standing approval.

The two high-priority issues are tied together: `list_unread` does an unbounded sequential N+1 pattern (`conversations.info` per channel) with no partial-result signal. Slack's tier for `conversations.info` is ~50 req/min, and the action currently fails the entire response on a single 429, so for a user in >50 conversations the feature is effectively broken. Compounding that, if `maxUserConversationPages` (50 pages × 200 = 10k) is hit, the result silently truncates with no `truncated` / `next_cursor` field, so an agent asking "any unreads?" can be given a false-negative answer.

The medium-priority item is `slack.mark_read`'s standing-approval template with wildcards on both `channel_id` and `ts` — "silently clear the user's unread badge in any conversation, forever" is a more sensitive capability than "read a list." Worth a tighter scope.

### What Looks Good
- Shared `fetchConversationInfo` is a nice abstraction — clean separation of types and helpers.
- `truncatePreviewText` correctly uses `utf8.RuneCountInString` and operates on `[]rune`, avoiding the classic byte-truncation emoji-splitting bug; the 250-`a` test asserts exactly the 201-rune post-ellipsis length.
- `mark_read` correctly runs `verifyChannelAccess` before mutating state, and the manifest + validator both make `ts` a required parameter rather than defaulting to "now."
- `ResolveResourceDetails` wiring for `slack.mark_read` reuses `resolveChannel` via the `channelOnly` marshal — consistent with the existing `search_messages` pattern.
- Tests cover the failure surfaces that matter (validation errors, rate-limit classification, pagination continuation).

---

---

chiedoclaude:
## Code Review Complete

**PR:** #1011 — Slack: list_unread and mark_read for agent unread state (#1010)
**Author:** @chiedocursor
**Rounds completed:** 1/6
**Reason for stopping:** PR merged before Round 2 wait elapsed

### Outstanding Issues

The PR merged with two **High**-severity findings unresolved. Flagging them here so they can be triaged as follow-up work rather than silently deferred:

1. **N+1 `conversations.info` fragility in `slack.list_unread`** (`connectors/slack/list_unread.go:77`). Sequential per-channel `conversations.info` against Slack's ~50 req/min tier means any user in more than ~50 conversations is likely to trip a rate limit mid-enumeration, and the current code returns the whole action as an error — none of the unreads already collected are surfaced. Suggested fixes: (a) swallow per-channel transient errors and return a `truncated: true` + partial result, (b) bounded parallelism (4–8 workers) for the info calls, and/or (c) add a `max_results` short-circuit so "do I have any unreads?" can stop scanning early.

2. **Silent pagination truncation at 10k channels** (`connectors/slack/list_unread.go:97`). When `maxUserConversationPages` is hit, the loop `break`s and the response has no `truncated` / `next_cursor` field, so an agent gets `unread_channels: []` and confidently tells the user "you're all caught up" even though pages 51+ were never inspected. Thread a `truncated` flag and cursor through `listUnreadResult`, and surface the limit in the README and manifest description.

### Follow-up Issue

Opened #1012 to track the non-blocking Medium/Low recommendations (standing-approval scope on `mark_read`, `conversations.info` caller consolidation, helper de-duplication).

### Review History
| 1 | 5 | 0 critical, 2 high, 2 medium, 1 low — PR merged before further rounds |

### Overall Assessment

The action shape, wiring, and tests are solid — idiomatic Go, correct reuse of `doGet` / `parseAndValidate` / `verifyChannelAccess`, and meaningful coverage of the key branches including UTF-8-safe preview truncation and rate-limit classification. The `fetchConversationInfo` helper is a nice consolidation (with the caveat that two existing callers still need to be migrated onto it).

The two open concerns — N+1 rate-limit fragility and silent pagination truncation in `list_unread` — are both about how this behaves at workspace scale. They are production-relevant but not crash-level, so shipping first and hardening via #1012 is a reasonable trade-off; just don't let the follow-ups sit. The `mark_read` standing-approval scope is more of a trust/UX call than a bug and is worth revisiting in that issue.

---

---

PR #1017, opened 2026-04-22
OAuth refresh: concurrent rotation race + multi-instance tests (closes #1014)

### LINKED ISSUE #1014: Harden OAuth refresh under multi-instance: test coverage + rotation race
## Context

A recent audit of OAuth refresh under multi-instance (multiple connections per user+provider) found the core plumbing is correctly scoped to individual `oauth_connections.id` — refresh works per-instance. But the audit turned up real gaps worth addressing in a follow-up. PR #1013 handled the separate routing and scope bugs; this issue covers the remaining refresh-side work.

Related files:
- `background_oauth.go`
- `api/connector_execution.go:383-540`
- `db/oauth_connections.go:100-145, 255-310`
- `oauth/refresh.go`
- `oauth/refresh_test.go`
- `db/oauth_connections_refresh_test.go`

## Phase 1 — Test coverage + cleanup (parallelizable with Phase 2)

Low-risk additive work. Safe to ship on its own.

- [ ] **Add a multi-instance refresh test.** Create two `oauth_connections` rows for the same `user_id` and `provider` (e.g. two Slack workspaces), with different refresh tokens, different vault secrets, and overlapping expiry windows inside the background job's 15-minute horizon. Assert that both refreshes run, each updates its own connection ID and only its own vault secrets, and neither cross-contaminates the other's `refresh_token_vault_id` or `access_token_vault_id`. Belongs in `background_oauth_test.go` (if it exists) or alongside `db/oauth_connections_refresh_test.go`.
- [ ] **Add an on-demand-refresh multi-instance test.** Call `credentialsFromOAuthConnection` serially on two different connections for the same user+provider, both near expiry, and assert each returns credentials derived from its own refreshed vault secret.
- [ ] **Delete dead code.** `db/oauth_connections.go:119-145` — `GetOAuthConnectionsByProvider` (plural) has zero callers. Remove it.
- [ ] **Fix `GetOAuthConnectionByProvider` ambiguity.** `db/oauth_connections.go:100-117` is only called from tests but returns whichever row Postgres finds first when multiple connections match. Add `ORDER BY created_at DESC LIMIT 1` so test behavior is deterministic. Update the doc comment to note it returns the most recent connection for the given provider.

**Suggested PR scope:** all four checkboxes in one PR.

## Phase 2 — Refresh-token-rotation race (can run in parallel with Phase 1)

Higher-risk behavioral change to the refresh hot path. Should be its own PR for easy revert.

### The problem

`api/connector_execution.go:442-445` has this comment:

> if the background refresh job refreshes the same connection concurrently, both paths produce valid tokens. The last writer wins in the DB, and the other's vault secrets become orphaned. This is safe (no data loss or auth failure)…

That's only true for providers that don't rotate refresh tokens. **Google, Microsoft, and Salesforce (with rotation enabled) invalidate the old refresh token the moment a new one is issued.** Sequence:

1. Background job reads connection X, starts refresh with refresh_token A.
2. User request triggers on-demand refresh on connection X, also using refresh_token A (DB hasn't been updated yet).
3. Background job succeeds → provider issues refresh_token B, invalidates A.
4. On-demand call reaches the provider with now-invalid refresh_token A → returns `invalid_grant`.
5. On-demand path (`api/connector_execution.go:484-491`) flips status to `needs_reauth`.
6. A healthy connection that was just successfully refreshed is now marked offline.

Multi-instance doesn't make this worse per-connection, but Nx more connections means Nx more absolute chances for any one user to hit the race.

### Recommended fix

**Re-read the connection row before flipping status to `needs_reauth`.** If `token_expiry` has moved forward or `updated_at` is newer than when we read the row, another path refreshed successfully in the meantime — don't flip status, just reload and retry the read from the (now-updated) vault secret.

- [ ] Add a helper `reloadAndMaybeSkipStatusFlip(ctx, db, conn, lastKnownUpdatedAt) (*OAuthConnection, bool)` that re-reads the row and returns `(freshConn, true)` if it was updated after we started. Callers use the fresh connection and skip the status flip.
- [ ] Apply it at both `needs_reauth` flip sites in `api/connector_execution.go:457-464` and `:484-491`.
- [ ] Apply it at the same sites in `background_oauth.go:138-142` and `:152-157`.
- [ ] Add a test that simulates the race: start a refresh, mid-flight bump `updated_at` and `token_expiry` on the same row (as if another path succeeded), let the in-flight refresh fail with `invalid_grant`, and assert the status stays `active` instead of flipping to `needs_reauth`.

Alternative (not chosen): advisory lock via `pg_advisory_xact_lock(hashtext(conn.id))` around the refresh. Cleaner but requires a transaction boundary around an HTTP call and is more invasive. Retry-read-before-flip is a narrower defensive check.

**Suggested PR scope:** all four checkboxes in one PR.

## Non-goals

- Surfacing `credentials_ready` in the frontend connector settings UI. Tracked separately if/when needed.
- Advisory locks around refresh. Covered by the alternative note above; revisit only if the retry-read approach proves insufficient.

## Acceptance

- `make test-backend` passes with the new tests.
- No regression in refresh behavior for providers without rotation.
- Healthy connections no longer flip to `needs_reauth` when a concurrent refresh succeeds.

https://claude.ai/code/session_01H43oDUXF1eZqAeF9Fi2KBK

chiedocursor:
## Summary

Implements [issue #1014](https://github.com/supersuit-tech/permission-slip/issues/1014): Phase 1 (tests + DB cleanup) and Phase 2 (reload-before-`needs_reauth` on refresh failure).

### Phase 1
- **Multi-instance background refresh test** (`background_oauth_test.go`): two Slack connections for the same user+provider, overlapping expiry in the 15m horizon; both refresh; vault IDs and plaintext access tokens stay distinct.
- **On-demand multi-instance test** (`api/connector_execution_test.go`): `credentialsFromOAuthConnection` called serially for two connections; each gets credentials from its own refresh path.
- **Removed** dead `GetOAuthConnectionsByProvider`.
- **`GetOAuthConnectionByProvider`**: `ORDER BY created_at DESC LIMIT 1` + doc comment.

### Phase 2
- **`db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded`**: re-reads the row; if `updated_at` is newer than the snapshot or `token_expiry` moved forward (or expiry appeared when absent), treat as concurrent success and **do not** flip to `needs_reauth`.
- Wired into **`refreshOAuthConnection`** (nil refresh token path + failed `RefreshTokens`) and **`refreshSingleConnection`** in `background_oauth.go`.
- **Race test**: simulated concurrent DB bump + `invalid_grant` from token endpoint; connection stays `active`. HTTP handler does not touch `pgx.Tx` concurrently; a helper goroutine runs the UPDATE on the test goroutine.

Closes #1014

## Test plan

- [ ] `make test-backend` (CI) — **not run locally**: sandbox Go version/module resolution cannot execute full `go test` per project notes; rely on CI for `make test-backend`.

## Self-review

- Concurrent refresh detection uses `updated_at` and `token_expiry` vs snapshot; covers rotation races where expiry moves forward.
- Nil→non-nil `token_expiry` treated as success path to avoid false `needs_reauth` if only expiry was backfilled.
- Invalid-grant tests coordinate DB updates off the httptest server goroutine to avoid concurrent use of the test transaction.

<div><a href="https://cursor.com/agents/bc-b38691fe-04f1-44b2-8639-88da5fdc5807"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b38691fe-04f1-44b2-8639-88da5fdc5807"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/api/connector_execution.go
+++ b/api/connector_execution.go
@@ -448,7 +448,7 @@ func credentialsFromOAuthConnection(ctx context.Context, deps *Deps, conn *db.OA
 // the old refresh token; the losing path may see invalid_grant. We re-read the row
 // before marking needs_reauth so we do not flip a healthy connection offline.
 func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConnection, providerID string) error {
-	lastKnownUpdatedAt := conn.UpdatedAt
+	lastKnownAccessVaultID := conn.AccessTokenVaultID
 	lastKnownTokenExpiry := conn.TokenExpiry
 
 	provider, ok := deps.OAuthProviders.Get(providerID)
@@ -460,13 +460,13 @@ func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConne
 	}
 
 	if conn.RefreshTokenVaultID == nil {
-		fresh, skipFlip, err := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownUpdatedAt, lastKnownTokenExpiry)
+		fresh, skipFlip, err := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownAccessVaultID, lastKnownTokenExpiry)
 		if err != nil {
 			return err
 		}
-		if skipFlip && fresh != nil && fresh.RefreshTokenVaultID != nil {
+		if skipFlip && fresh != nil && fresh.Status == db.OAuthStatusActive && fresh.RefreshTokenVaultID != nil {
 			*conn = *fresh
-			lastKnownUpdatedAt = conn.UpdatedAt
+			lastKnownAccessVaultID = conn.AccessTokenVaultID
 			lastKnownTokenExpiry = conn.TokenExpiry
 		} else {
 			// No refresh token — can't refresh. Mark as needs_reauth.
@@ -497,11 +497,11 @@ func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConne
 	if err != nil {
 		// Refresh failed (token revoked, expired, invalid_grant after rotation, etc.).
 		// Re-read: another goroutine may have refreshed successfully and invalidated our refresh token.
-		fresh, skipFlip, reloadErr := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownUpdatedAt, lastKnownTokenExpiry)
+		fresh, skipFlip, reloadErr := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownAccessVaultID, lastKnownTokenExpiry)
 		if reloadErr != nil {
 			return reloadErr
 		}
-		if skipFlip && fresh != nil {
+		if skipFlip && fresh != nil && fresh.Status == db.OAuthStatusActive {
 			*conn = *fresh
 			return nil
 		}
--- a/api/connector_execution.go
+++ b/api/connector_execution.go
@@ -443,7 +443,14 @@ func credentialsFromOAuthConnection(ctx context.Context, deps *Deps, conn *db.OA
 // both paths produce valid tokens. The last writer wins in the DB, and the other's
 // vault secrets become orphaned. This is safe (no data loss or auth failure) but
 // may leave unused vault entries until the connection is deleted or re-authorized.
+//
+// When a provider rotates refresh tokens, a concurrent successful refresh invalidates
+// the old refresh token; the losing path may see invalid_grant. We re-read the row
+// before marking needs_reauth so we do not flip a healthy connection offline.
 func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConnection, providerID string) error {
+	lastKnownUpdatedAt := conn.UpdatedAt
+	lastKnownTokenExpiry := conn.TokenExpiry
+
 	provider, ok := deps.OAuthProviders.Get(providerID)
 	if !ok {
 		return &connectors.OAuthRefreshError{
@@ -453,14 +460,24 @@ func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConne
 	}
 
 	if conn.RefreshTokenVaultID == nil {
-		// No refresh token — can't refresh. Mark as needs_reauth.
-		if statusErr := db.UpdateOAuthConnectionStatus(ctx, deps.DB, conn.ID, conn.UserID, db.OAuthStatusNeedsReauth); statusErr != nil {
-			log.Printf("failed to update OAuth connection status to needs_reauth: %v", statusErr)
-			CaptureError(ctx, statusErr)
+		fresh, skipFlip, err := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownUpdatedAt, lastKnownTokenExpiry)
+		if err != nil {
+			return err
 		}
-		return &connectors.OAuthRefreshError{
-			Provider: providerID,
-			Message:  "token expired and no refresh token available — user must re-authorize",
+		if skipFlip && fresh != nil && fresh.RefreshTokenVaultID != nil {
+			*conn = *fresh
+			lastKnownUpdatedAt = conn.UpdatedAt
+			lastKnownTokenExpiry = conn.TokenExpiry
+		} else {
+			// No refresh token — can't refresh. Mark as needs_reauth.
+			if statusErr := db.UpdateOAuthConnectionStatus(ctx, deps.DB, conn.ID, conn.UserID, db.OAuthStatusNeedsReauth); statusErr != nil {
+				log.Printf("failed to update OAuth connection status to needs_reauth: %v", statusErr)
+				CaptureError(ctx, statusErr)
+			}
+			return &connectors.OAuthRefreshError{
+				Provider: providerID,
+				Message:  "token expired and no refresh token available — user must re-authorize",
+			}
 		}
 	}
 
@@ -478,8 +495,16 @@ func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConne
 	defer refreshCancel()
 	result, err := oauth.RefreshTokens(refreshCtx, provider, string(refreshTokenBytes))
 	if err != nil {
-		// Refresh failed (token revoked, expired, etc.). Mark as needs_reauth.
-		// Log the full error server-side for debugging; return a sanitized message to the caller.
+		// Refresh failed (token revoked, expired, invalid_grant after rotation, etc.).
+		// Re-read: another goroutine may have refreshed successfully and invalidated our refresh token.
+		fresh, skipFlip, reloadErr := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownUpdatedAt, lastKnownTokenExpiry)
+		if reloadErr != nil {
+			return reloadErr
+		}
+		if skipFlip && fresh != nil {
+			*conn = *fresh
+			return nil
+		}
 		log.Printf("oauth refresh failed for provider %q connection %s: %v", providerID, conn.ID, err)
 		if statusErr := db.UpdateOAuthConnectionStatus(ctx, deps.DB, conn.ID, conn.UserID, db.OAuthStatusNeedsReauth); statusErr != nil {
 			log.Printf("failed to update OAuth connection status to needs_reauth: %v", statusErr)
--- a/api/connector_execution_test.go
+++ b/api/connector_execution_test.go
@@ -596,12 +596,13 @@ func TestExecuteConnectorAction_OAuthPath_RefreshInvalidGrantConcurrentSuccessDo
 	// from the httptest server goroutine (tx is not safe for concurrent use).
 	bumpCh := make(chan struct{})
 	bumpDone := make(chan struct{})
+	var bumpErr error
 	go func() {
 		defer close(bumpDone)
 		<-bumpCh
 		newAccessID := "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"
-		testhelper.MustExec(t, tx,
-			`UPDATE oauth_connections SET
+		_, bumpErr = tx.Exec(t.Context(), `
+			UPDATE oauth_connections SET
 				token_expiry = now() + interval '2 hours',
 				updated_at = now(),
 				access_token_vault_id = $1,
@@ -614,6 +615,10 @@ func TestExecuteConnectorAction_OAuthPath_RefreshInvalidGrantConcurrentSuccessDo
 	tokenSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		close(bumpCh)
 		<-bumpDone
+		if bumpErr != nil {
+			http.Error(w, bumpErr.Error(), http.StatusInternalServerError)
+			return
+		}
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusBadRequest)
 		json.NewEncoder(w).Encode(map[string]any{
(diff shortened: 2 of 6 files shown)


db/oauth_connections.go
```diff
@@ -116,31 +119,34 @@ func GetOAuthConnectionByProvider(ctx context.Context, db DBTX, userID, provider
...
-		c, err := scanOAuthConnection(rows.Scan)
-		if err != nil {
-			return nil, err
-		}
-		conns = append(conns, c)
+	if fresh == nil || fresh.UserID != userID {
+		return nil, false, nil
 	}
-	return conns, rows.Err()
+	if fresh.UpdatedAt.After(lastKnownUpdatedAt) {
```
chiedoclaude: **[High · Correctness]** `UpdatedAt.After(lastKnownUpdatedAt)` fires on *any* row update, not only successful concurrent refreshes. `UpdateOAuthConnectionStatus` also sets `updated_at = now()` — so if a concurrent refresh path **also failed** and flipped the row to `needs_reauth`, this returns `skipFlip=true` with a stale `fresh` row whose `Status == needs_reauth` and whose `AccessTokenVaultID` points at the pre-refresh secret.

**Why it matters:** the callers in `api/connector_execution.go:504` and `background_oauth.go:167` then `return nil` as if the refresh succeeded. `credentialsFromOAuthConnection` proceeds to read the stale access-token vault entry and hands it to the connector. The downstream API call will 401, the user sees a confusing opaque error instead of the correct `OAuthRefreshError`, and two concurrent failing refreshes now silently produce bogus credentials — a regression from pre-PR behaviour where both paths would flip `needs_reauth` and return an error.

**Fix:** require `fresh.Status == OAuthStatusActive` before declaring a concurrent success. Even better, use the refresh-specific signals (new access-token vault ID or advanced `token_expiry`) rather than plain `updated_at`:

```go
if fresh.Status != OAuthStatusActive {
    return fresh, false, nil
}
if fresh.UpdatedAt.After(lastKnownUpdatedAt) {
    return fresh, true, nil
}
// ...existing token_expiry checks...
```

Then add a regression test: two concurrent failing refreshes, assert the second caller still gets an `OAuthRefreshError` instead of stale creds.

api/connector_execution.go
```diff
@@ -478,8 +495,16 @@ func refreshOAuthConnection(ctx context.Context, deps *Deps, conn *db.OAuthConne
...
 	if err != nil {
-		// Refresh failed (token revoked, expired, etc.). Mark as needs_reauth.
-		// Log the full error server-side for debugging; return a sanitized message to the caller.
+		// Refresh failed (token revoked, expired, invalid_grant after rotation, etc.).
+		// Re-read: another goroutine may have refreshed successfully and invalidated our refresh token.
+		fresh, skipFlip, reloadErr := db.ReloadOAuthConnectionIfConcurrentRefreshSucceeded(ctx, deps.DB, conn.ID, conn.UserID, lastKnownUpdatedAt, lastKnownTokenExpiry)
+		if reloadErr != nil {
+			return reloadErr
+		}
+		if skipFlip && fresh != nil {
```
chiedoclaude: **[High · Correctness]** Companion to the `db/oauth_connections.go:139` finding. Even once the reload helper is tightened, consider adding a belt-and-braces `fresh.Status == OAuthStatusActive` check here before `return nil` — the caller (`credentialsFromOAuthConnection`) does not re-validate `conn.Status` after `refreshOAuthConnection` returns, so any path where this returns `nil` with a non-active status will serve credentials from a connection the system considers offline.

```go
if skipFlip && fresh != nil && fresh.Status == db.OAuthStatusActive {
    *conn = *fresh
    return nil
}
```

Same pattern should be mirrored in `background_oauth.go:167`.

api/connector_execution_test.go
```diff
@@ -581,6 +581,210 @@ func TestExecuteConnectorAction_OAuthPath_RefreshesExpiredToken(t *testing.T) {
...
+	testhelper.InsertUser(t, tx, uid, "u_"+uid[:8])
+
+	// Run the concurrent DB bump on a dedicated goroutine so we never touch pgx.Tx
+	// from the httptest server goroutine (tx is not safe for concurrent use).
+	bumpCh := make(chan struct{})
+	bumpDone := make(chan struct{})
+	go func() {
+		defer close(bumpDone)
+		<-bumpCh
+		newAccessID := "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"
```
chiedoclaude, line 603: **[Medium · Testing]** `testhelper.MustExec` calls `t.Fatalf` on error, and it's being invoked here from a goroutine other than the test goroutine. Per Go's testing contract, `Fatal`/`Fatalf`/`FailNow` must only be called from the test goroutine — `FailNow` calls `runtime.Goexit()`, which terminates *only the calling goroutine*, leaving the main test to continue with stale/unclear state and potentially hang on `bumpDone` or report a misleading failure.

**Why it matters:** if the `UPDATE` ever fails (migration drift, constraint change, pool issue), the test becomes unreliable and the failure mode is a confusing timeout or misattributed assertion rather than a clean fatal.

**Fix:** capture the error in the goroutine and assert it on the main test goroutine, e.g.:

```go
var bumpErr error
go func() {
    defer close(bumpDone)
    <-bumpCh
    _, bumpErr = tx.Exec(context.Background(), `UPDATE ...`, ...)
}()
// later, on the main goroutine:
<-bumpDone
if bumpErr != nil { t.Fatalf("bump: %v", bumpErr) }
```

The same issue exists in `background_oauth_test.go:202`.

background_oauth_test.go
```diff
@@ -70,3 +78,189 @@ func TestStartOAuthRefresh_StopsOnContextCancel(t *testing.T) {
...
+	uid := testhelper.GenerateUID(t)
+	testhelper.InsertUser(t, tx, uid, "u_"+uid[:8])
+
+	bumpCh := make(chan struct{})
+	bumpDone := make(chan struct{})
+	go func() {
+		defer close(bumpDone)
+		<-bumpCh
+		newAccessID := "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"
+		testhelper.MustExec(t, tx,
```
chiedoclaude: **[Medium · Testing]** Same concern as `api/connector_execution_test.go:602` — `testhelper.MustExec` (which calls `t.Fatalf`) is invoked from a non-test goroutine. Switch to capturing the error and asserting it from the main goroutine after `<-bumpDone`. Similarly, the `t.Errorf("parse form: %v", err)` at line 91 inside the httptest handler is technically also outside the test goroutine — `t.Errorf` is a bit safer than `t.Fatalf` but still documented as test-goroutine-only.

db/oauth_connections.go
```diff
@@ -97,13 +97,16 @@ func ListOAuthConnectionsByUser(ctx context.Context, db DBTX, userID string) ([]
...
 
-// GetOAuthConnectionByProvider returns the OAuth connection for a given user and provider.
-// Returns nil if no connection exists.
+// GetOAuthConnectionByProvider returns the most recent OAuth connection for a given user and
+// provider (ORDER BY created_at DESC LIMIT 1). When multiple connections exist for the same
+// user+provider, this is deterministic. Returns nil if no connection exists.
 func GetOAuthConnectionByProvider(ctx context.Context, db DBTX, userID, provider string) (*OAuthConnection, error) {
 	row := db.QueryRow(ctx, `
 		SELECT `+oauthConnectionColumns+`
 		FROM oauth_connections
```
chiedoclaude, line 106: **[Low · Correctness]** `ORDER BY created_at DESC LIMIT 1` is deterministic only when `created_at` values are distinct. Postgres `timestamptz` has microsecond precision, so two rows inserted in the same transaction or same tick can tie — which is exactly the multi-instance scenario this PR is hardening. Add a secondary sort on a strictly-ordered column for full determinism:

```sql
ORDER BY created_at DESC, id DESC
LIMIT 1
```

db/oauth_connections.go
```diff
@@ -20,29 +20,29 @@ const (
...
+	UserID              string
+	Provider            string
+	AccessTokenVaultID  string
+	RefreshTokenVaultID *string
+	Scopes              []string
+	TokenExpiry         *time.Time
+	Status              string
+	ExtraData           json.RawMessage // provider-specific data (e.g. Salesforce instance_url)
+	CreatedAt           time.Time
+	UpdatedAt           time.Time
```
chiedoclaude, line 33: **[Low · Maintainability]** The struct-field whitespace re-alignment on `OAuthConnection` and `CreateOAuthConnectionParams` is a cosmetic reformat of a hot file. Per the repo's merge-conflict guidance (CLAUDE.md → "Avoid reformatting or reordering existing code"), this touches unrelated lines and will conflict with any concurrent branch editing these structs. Recommend reverting the whitespace-only hunk — it's unrelated to the bug fix and adds merge risk for no behavioural win.

chiedoclaude (commented):
## Review Round 1/6

**Findings:** 5 (0 critical, 1 high, 2 medium, 2 low)

The PR delivers a well-structured fix for issue #1014 — the multi-instance tests are a genuinely useful addition and the `ReloadOAuthConnectionIfConcurrentRefreshSucceeded` helper narrows the rotation-race window as intended. One substantive concern: the reload helper treats *any* `updated_at` bump as evidence of a successful concurrent refresh, which misclassifies the case where the concurrent path also failed and flipped to `needs_reauth` — the losing path would then return `nil` to the caller and serve stale credentials. There are also two test-reliability nits (calling `t.Fatalf` via `testhelper.MustExec` from non-test goroutines) and a couple of low-severity cleanups.

### What Looks Good
- Multi-instance refresh tests (`TestRefreshSingleConnection_MultiInstanceNoCrossContamination`, `TestCredentialsFromOAuthConnection_MultiInstanceSameProvider`) clearly assert isolation of vault IDs across connections — exactly what Phase 1 of the issue asked for.
- The `ReloadOAuthConnectionIfConcurrentRefreshSucceeded` docstring explains the race model clearly, and the defense-in-depth `UserID` check is a nice touch.
- Dropping the unused `GetOAuthConnectionsByProvider` and tightening `GetOAuthConnectionByProvider` with `LIMIT 1` is solid cleanup.
- Channel-based sequencing between the httptest handler and the DB-bump goroutine is the right shape for the race test.

---

db/oauth_connections.go
chiedocursor: Implemented in 8e57c1a4: reload helper requires active status plus refresh-specific signals (access vault id changed vs snapshot, or token_expiry advanced/backfilled). Callers pass lastKnownAccessVaultID and gate skip on active. GetOAuthConnectionByProvider uses ORDER BY created_at DESC, id DESC. Reverted struct whitespace-only reformat. Race tests capture tx.Exec errors on bump goroutine; httptest uses http.Error for parse failures. Added regression tests for concurrent needs_reauth + invalid_grant.

api/connector_execution.go
chiedocursor: Implemented in 8e57c1a4: reload helper requires active status plus refresh-specific signals (access vault id changed vs snapshot, or token_expiry advanced/backfilled). Callers pass lastKnownAccessVaultID and gate skip on active. GetOAuthConnectionByProvider uses ORDER BY created_at DESC, id DESC. Reverted struct whitespace-only reformat. Race tests capture tx.Exec errors on bump goroutine; httptest uses http.Error for parse failures. Added regression tests for concurrent needs_reauth + invalid_grant.

api/connector_execution_test.go
chiedocursor: Implemented in 8e57c1a4: reload helper requires active status plus refresh-specific signals (access vault id changed vs snapshot, or token_expiry advanced/backfilled). Callers pass lastKnownAccessVaultID and gate skip on active. GetOAuthConnectionByProvider uses ORDER BY created_at DESC, id DESC. Reverted struct whitespace-only reformat. Race tests capture tx.Exec errors on bump goroutine; httptest uses http.Error for parse failures. Added regression tests for concurrent needs_reauth + invalid_grant.

background_oauth_test.go
chiedocursor: Implemented in 8e57c1a4: reload helper requires active status plus refresh-specific signals (access vault id changed vs snapshot, or token_expiry advanced/backfilled). Callers pass lastKnownAccessVaultID and gate skip on active. GetOAuthConnectionByProvider uses ORDER BY created_at DESC, id DESC. Reverted struct whitespace-only reformat. Race tests capture tx.Exec errors on bump goroutine; httptest uses http.Error for parse failures. Added regression tests for concurrent needs_reauth + invalid_grant.

db/oauth_connections.go
chiedocursor: Implemented in 8e57c1a4: reload helper requires active status plus refresh-specific signals (access vault id changed vs snapshot, or token_expiry advanced/backfilled). Callers pass lastKnownAccessVaultID and gate skip on active. GetOAuthConnectionByProvider uses ORDER BY created_at DESC, id DESC. Reverted struct whitespace-only reformat. Race tests capture tx.Exec errors on bump goroutine; httptest uses http.Error for parse failures. Added regression tests for concurrent needs_reauth + invalid_grant.

db/oauth_connections.go
chiedocursor: Implemented in 8e57c1a4: reload helper requires active status plus refresh-specific signals (access vault id changed vs snapshot, or token_expiry advanced/backfilled). Callers pass lastKnownAccessVaultID and gate skip on active. GetOAuthConnectionByProvider uses ORDER BY created_at DESC, id DESC. Reverted struct whitespace-only reformat. Race tests capture tx.Exec errors on bump goroutine; httptest uses http.Error for parse failures. Added regression tests for concurrent needs_reauth + invalid_grant.

chiedocursor:
## 🤖 Watch Session Summary

### Changes Made
- **`Resolved PR review threads after drain pass`** (`8e57c1a4`) — GraphQL resolveReviewThread for 6 thread IDs from watch-poll drain

### Decision Log

#### ✅ Implemented
- chiedoclaude asked for reload helper: active + vault/expiry signals → implemented in `8e57c1a4`
- chiedoclaude asked for caller active check before return nil → implemented in `8e57c1a4`
- chiedoclaude asked for race test: no MustExec from bump goroutine → implemented in `8e57c1a4`
- chiedoclaude asked for background_oauth_test bump goroutine + parseForm → implemented in `8e57c1a4`
- chiedoclaude asked for ORDER BY id DESC tie-break → implemented in `8e57c1a4`
- chiedoclaude asked for revert struct whitespace reformat → implemented in `8e57c1a4`
- chiedoclaude asked for review round 1 summary (all findings addressed) → implemented in `8e57c1a4`

---
*Watch session ended after 0 minutes of inactivity. Processed 7 comments across 2 poll cycles.*

PR #1106, opened 2026-04-25
fix: unblock local dev first-run

jayfallon:
## Summary

Walking a fresh checkout through the local dev quickstart in `docs/development.md` hits several blockers before the first agent can register and run an action. This PR fixes the ones that have a clear correct answer; one related blocker (the GitHub connector seed) is intentionally left for a deliberate decision (see **Known issue not fixed here** below).

Each fix below is independent and small; happy to split if a reviewer prefers.

### Fixes

- **`cli/src/auth/keys.ts`** — Parse OpenSSH-format ed25519 private keys (the format `ssh-keygen` produces by default). The previous comment claimed Node's `crypto.createPrivateKey` accepts both PKCS8 and OpenSSH PEM — it does not, and `permission-slip register` failed with `error:1E08010C:DECODER routines::unsupported`. The fix manually parses the OpenSSH private-key blob to extract the 32-byte ed25519 seed + public key, then constructs the `KeyObject` from a JWK. PKCS8 PEM keys (the existing `writeOpenSSHPrivateKey` fallback path) still load via the same code path as before. Existing `cli/tests/keys.test.ts` continues to pass.

- **`supabase/config.toml`** — Bump `auth.email.otp_length` from `6` to `8`. The frontend's `OtpStep` and the backend's `shared/validation.go` both enforce `emailOtpCode.length = 8` from `shared/validation.json` — Supabase shipping 6-digit codes meant the OTP sign-in screen's Verify button stayed disabled forever and `emailOtpCode.length` was effectively unenforced.

- **`frontend/src/auth/dev.ts`** — ⚠️ **Workaround, flagged for follow-up.** The "Dev: Auto-fill code from Mailpit" button extracted the OTP with a regex hardcoded to `\d{6}`. Once `validation.json` says `length: 8`, that regex returns nothing and auto-fill silently fails. We patched the regex to read the length from `validation.json`. Without this, the local dev sign-in flow is impossible to auto-complete; with it, anyone who later changes `emailOtpCode.length` is fine. The cleaner long-term fix is probably to source OTP length from a single shared place at build time so this regex doesn't need a runtime dynamic-import.

- **`frontend/vite.config.ts`** — Re-add the frontend project root to `server.fs.allow`. Commit 2952d417 (`Narrow free plan type and tighten Vite fs.allow scope`) reduced the allow-list from `..` to `../config` only. Setting `fs.allow` *replaces* Vite's defaults (which include the project root), so `frontend/index.html` and `src/main.tsx` were no longer servable in dev — the Vite log spammed `outside of Vite serving allow list` for every request. The fix preserves the security intent: the new entries are `__dirname` (frontend root) + `../config` only, so Go source / migrations / `.env` outside `frontend/` stay denied.

- **`frontend/src/pages/agents/connectors/AddCredentialDialog.tsx`** + **`SetupConnectorCredentialsDialog.tsx`** — Per-service field overrides for connectors with `auth_type: "custom"`. Every custom-auth connector previously rendered a single field labeled **"API Key"** stored under the key `api_key`. PostgreSQL credentials silently went into the wrong slot (`api_key` instead of `connection_string`); the connector's `Validate` later failed with `missing required credential: connection_string`. This PR adds a small lookup table for `postgres`/`mysql`/`snowflake` (all want `connection_string`) and `redis` (wants `url`) so the dialog labels and stores them correctly. **This is a stopgap** — the proper fix is to extend `connectors.ManifestCredential` with field schemas (key + label + placeholder) and have the dialog render generically. Filing a follow-up issue would unstick the next custom-auth connector someone adds.

### Known issue not fixed here

The GitHub connector's manifest declares both an OAuth2 (`github`) and an API-key (`github_pat`) credential. Migration `20260324120723_prevent_mixed_auth_type_connectors.sql` installs a trigger that rejects any connector with both, so the seed transaction fails and GitHub never gets a `connector_required_credentials` row — making the connector invisible in the dashboard. The fix is a product/architecture call (drop PAT support, split into two connectors, or make the seeder tolerant of the trigger), so I left it alone here. Workaround during testing: temporarily delete the `github_pat` block from the manifest.

## Test plan

- [x] `cd cli && npm run build` clean
- [x] `cd cli && npm test -- keys.test` (7/7 pass — covers OpenSSH key path)
- [x] `cd frontend && npx tsc --noEmit` clean
- [x] `cd frontend && npx vitest run src/auth/__tests__/OtpStep.test.tsx` (12/12 pass)
- [ ] Backend (`go test ./...`): not run locally — these changes don't touch Go, but CI should confirm
- [ ] Manual: fresh `supabase db reset` → sign up via Mailpit → "Dev: Auto-fill code from Mailpit" populates 8-digit code → submit succeeds
- [ ] Manual: build CLI → `register --invite-code` succeeds with no `DECODER` error → `verify --code` succeeds
- [ ] Manual: connect PostgreSQL → dialog labels the field "Connection String" → execute a `postgres.query` end-to-end

### DIFF
--- a/cli/src/auth/keys.ts
+++ b/cli/src/auth/keys.ts
@@ -50,6 +50,11 @@ export function readPublicKey(): string {
 
 /**
  * Loads the private key from disk as a Node KeyObject.
+ *
+ * Supports both PKCS8 PEM (`-----BEGIN PRIVATE KEY-----`, what Node natively
+ * understands) and the OpenSSH native format (`-----BEGIN OPENSSH PRIVATE KEY-----`,
+ * what `ssh-keygen` produces by default). For OpenSSH keys, the 32-byte ed25519
+ * seed is extracted manually and converted to a JWK before construction.
  */
 export function loadPrivateKey(): crypto.KeyObject {
   if (!fs.existsSync(PRIVATE_KEY_FILE)) {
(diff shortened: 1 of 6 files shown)


chiedo:
Looks great man! Opened follow issues for the things you called out. I've been running off prod for some time 😆

Will merge as is! But yeah if I'm so lucky to have more PRs from you in the future, multiple split up PRs is preferable. Beggars can't be choosers though!

chiedo:
🎉 first outside PR!

